PR #16, opened 2021-07-06
Update runme_FreeMoCap.py

kelly-phalen:
test to see pull requests


aaroncherian:
i don't like this change

runme_FreeMoCap.py
```diff
@@ -1,3 +1,4 @@
+#test
```
jonmatthis, line 1: this is a comment

PR #17, opened 2021-07-08
Aaronc fmc run me development

aaroncherian:
Merging the init freemocap runme and fmc_runme files into the main branch


freemocap/__init__.py
```diff
@@ -3,249 +3,7 @@
...
-    if stage <= 8:
-        print()
-        print("Starting Video Creation")
-        createvideo.createVideo(sesh)
-    else:
-        print("Skipping Video Creation")
-
-
-    f = 9
+def RunMe(sessionID,stage,useOpenPose,useMediaPipe,useDLC,debug):
```
jonmatthis: @aaroncherian - btw, you can set default values for functions like ths - 

def RunMe(
sessionID=None,
stage=1,
useOpenPose=True,
useMediaPipe=False,
useDLC=True,
debug=False
)

freemocap/fmc_runme.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+    # %% Stage Eight
+    if stage <= 8:
+        print()
+        print('Starting Video Creation')
+        createvideo.createVideo(sesh)
+    else:
+        print('Skipping Video Creation')
+
```
jonmatthis, line 203: Suggest adding this: 

```python
if __name__ == '__main__':
    Run()
```

To the bottom of the file, so this .py file can be run as a script (with default values, see previous comment)

jonmatthis (changes requested):
Hey look, we're like real software humans :O

freemocap/__init__.py
jonmatthis: if the user specifies any of these inputs, it will respect it, but if they leave it blank (i.e. `Run()`) it uses the default values

aaroncherian: got it - I added the default values to both the RunMe (from init) and Run (from fmc_runme), so that way whether you run freemocap.RunMe() or the if name = main 'Run()', you can use default values

freemocap/fmc_runme.py
aaroncherian: will save this for later: https://stackoverflow.com/questions/25888484/run-python-file-in-child-package-from-console

jonmatthis (approved):
Nice, looks good!

jonmatthis:
Done!

PR #24, opened 2021-07-19
Raw data to fileformat converter

jakobwelner:
Reworked tools for converting openpose raw data into a more accessible and self-contained format. You can now feed in your own mappings and parents when importing the data into the data handler object.
Maya tool simply needs to be fed a data handler object with at least one actor


jonmatthis:
Awesome! Can't wait to check it out!

@aaroncherian - could you give this a shot? Just follow the instructions at the top of `fmc_trackdata_handler.py`

PR #27, opened 2021-07-22
Fix issue 19

rudiejd:
Fix issue 19: https://github.com/jonmatthis/freemocap/issues/19


endurance:
The old code has been remove, so we're closing old issues and PRs related to the old code.

PR #29, opened 2021-07-22
Using newer Ubuntu 18.04 and fix tests

abitrolly:


abitrolly:
I give up for now. Something is happening with tests on GUI side and I don't see what,

aaroncherian:
Hey thanks for taking a look at this! I'll give it a shot and see if I can figure out what's going on with the framework issue. I ran into the issue way back at the start of this project, and it was something related to deeplabcut - and I just need to try and remember how exactly I solved it/if this is a similar issue

abitrolly:
@aaroncherian if it can be reproduced with minimal example, maybe reporting the issue to DeepLabCut would be easier?.

PR #43, opened 2021-08-16
Update READ.me with screen recording of install

kileyhartigan:
I simply made a screen recording of the install process and added a link to the vimeo video in the Installation section


endurance:
Thank you!

PR #50, opened 2021-09-01
.travis.yml: Run `pytest` in verbose mode to debug freeze

abitrolly:
`No output has been received in the last 10m0s.` is how the test are
currently failing, which started after installing `xvfb` in df74d3f . If
the attempt to thaw it out won't succeeded, the commit would need
to be reverted, because partial test result are better that no tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,44 @@
+  
+language: python
+python:
+  - 3.7
+cache:
+  directories:
+    - $HOME/.cache/pip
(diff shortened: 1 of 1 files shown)


abitrolly:
I don't see any builds firing. @jonmatthis @aaroncherian have you disabled Travis?

abitrolly:
The project run over the free quota allowed for OSS projects. Time to move to GitHub Actions.

abitrolly:
#96

PR #58, opened 2021-09-28
freemocap.wiki version 0

kileyhartigan:


kileyhartigan:
I know these instructions aren't really very professional, much more "personal-notes" style. So, let me know what changes you'd like me to make and I'll get working on them. 

One thing that my fastidious mind is irritated by is that the documentation_images/ folder contains ALL (80-ish) photos for my entire Mathis lab freemocap obsidian vault, not just for freemocap.wiki folder within that vault. However, the wiki only references a couple dozen of those 80-ish photos. So, functionally, it works, but it's messy. That's a limitation of image folder organization for obsidian, I can only have one images folder for each obsidian vault.

PR #86, opened 2022-01-03
Blender parented skeleton and baked animation

jallex:


jallex:
I just realized that the line I changed on fmc_runme.py might have to be changed to

output = subprocess.run([str(blenderPath), "--background", "--python", "fmc_blender.py", "--", str(sesh.dataArrayPath/'mediaPipeSkel_3d.npy'), str(sesh.dataArrayPath), str(sesh.sessionID)], capture_output=True, text=True, check=True)

I'm not totally sure if that change is necessary or not since I can't test this file on my side, but if it flags an error it might be that!

PR #88, opened 2022-01-12
Docker + FastAPI Refactor

endurance:
Our first cut of an API and using Docker to host it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 # Byte-compiled / optimized / DLL files
+.idea/*
 
 __pycache__/
 Data/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,40 @@
+# syntax=docker/dockerfile:1.2
+
+# Good base image to start from for most development
+FROM python:3.8-slim
+
+# Please remember, the base image we use /must be as small as possible/ for the best
+# production deployments. This is not optional.
+
+WORKDIR /workspace
+
+# Allows python to stream logs rather than buffer them for output.
+ENV PYTHONUNBUFFERED=1
+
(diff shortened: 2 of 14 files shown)


endurance:
Hey! Tell me more about your question, the "URL map" specifically. 

Here's a good reference for FastAPI's SDK
https://fastapi.tiangolo.com/tutorial/first-steps/

What we're setting up in this project is the beginning of the API that will service an Electron UI. We're going to use an MVC based design pattern where the "routers" are the "Controllers" and the pydantic models will provide the Model that back our objects for the API.

Jon Matthis, the owner of the project, hired me to help build out a couple of things and this is our first step. Just wanted to give some context.

abitrolly:
Yea. In Django https://docs.djangoproject.com/en/4.0/topics/http/urls/ URLs are usually contained in a single file `urls.py`, which also imports urls from other apps if needed. So it is easy to get an overview of what endpoints an application provides. In Ruby on Rails routes are also centralized in `config/routes.rb` https://guides.rubyonrails.org/routing.html#configuring-the-rails-router

Here to get an overview of different endpoints, one need to go to each handler.

endurance:
Oh okay, I see! You were referring centralized routing. FastAPI doesn't push users towards a centralized routing framework. FastAPI's routing pattern is more like ASP.NET MVC, Flask, Express or any other decorator based MVC-like API framework.

In this application, we'll be taking a conventional REST routing approach where the router folders will match the resource names exposed so it will be very obvious.

I haven't posted up the "CONTRIBUTING_GUIDLINES.md" file yet, but once we stabilize the repo, this will definitely be a topic covered for how open source devs can continue to contribute to this project.

PR #117, opened 2022-03-09
Endurance/dev branch

endurance:


.pre-commit-config.yaml
```diff
@@ -0,0 +1,10 @@
+repos:
```
endurance, line 1: pre commit hooks added to repo that ensure that all commits are formatted

PR #127, opened 2022-03-21
Jon/camera calibration

jonmatthis:
Building functionality for run-time, interactive camera calibration 

Restructured things so the `cv_camera_manager` context manager happens in a `SessionManager` object that manages calibration/board_detection/etc 

wound up changing a lot of variable names and whatnot


jonmatthis:
btw - the camera calibration functionality is currently turned off in the `__main__` function of `session_manager.py`

freemocap-ui/src/layout/Navigator.tsx
```diff
@@ -20,7 +20,7 @@ const categories = [
       },
       {
         id: "Board Detection",
-        route: "/board_detection"
+        route: "/charuco_board_detection"
```
endurance, line 23: when doing the renames, make sure to watch out for changes in the frontend code

src/cameras/capture/opencv_camera/frame_grabber.py
```diff
@@ -71,8 +71,8 @@ def _start_frame_loop(self, save_video=False):
         start = time.time()
         try:
             while self._is_capturing_frames:
-                success, image, timestamp = self._get_next_frame()
-                self._frame = FramePayload(success, image, timestamp)
+                success, image, timestamp, webcam_id = self._get_next_frame()
```
endurance, line 74: lets change this to:

```suggestion
                frame = self._get_next_frame()
                self._frame = frame
```

and then use `frame` downstream in the method if necessary.

src/cameras/capture/opencv_camera/opencv_camera.py
```diff
@@ -23,11 +26,26 @@ class WebcamConfig(BaseModel):
...
     save_video: bool = True
-    fourcc: str = "MP4V"
-    # fourcc: str = "MJPG"
+    # fourcc: str = "MP4V"
+    fourcc: str = "MJPG"
     base_save_video_dir = _get_home_dir()
 
 
+@dataclass
+class CameraCalibrationData:
```
endurance, line 35: move this class to another file

src/cameras/capture/opencv_camera/opencv_camera.py
```diff
@@ -23,11 +26,26 @@ class WebcamConfig(BaseModel):
...
+    image_width: int
+    image_height: int
+    lens_distortion_calibration_data: LensDistortionCalibrationData = None
+    camera_translation_relative_to_camera0: np.ndarray = np.zeros((3, 1))
```
endurance, line 39: what is "camera0"? how do we identify it?

src/cameras/capture/opencv_camera/opencv_camera.py
```diff
@@ -38,6 +56,23 @@ def __init__(self, config: WebcamConfig):
...
+
+        if self._camera_calibration_data is None and self.image_width is not None:
+            self._camera_calibration_data = CameraCalibrationData(self.image_width, self.image_height)
+            return self._camera_calibration_data
+
+        logger.warning(
+            'Can\'t request a cameras calibration info until the video capture is initialized and the image_width and image_height are known')
+
+    @camera_calibration_data.setter
+    def camera_calibration_data(self, camera_calibration_data: CameraCalibrationData):
```
endurance, line 74: this indicates that we would be setting calibration data from within the OpenCVCamera class - i think that breaks our current definition of what OCVCam does.

src/cameras/capture/opencv_camera/opencv_camera.py
```diff
@@ -108,15 +144,23 @@ def _create_thread(self):
...
-            frame_height=self.get_frame_height(),
+            frame_width=self.image_width,
+            frame_height=self.image_height,
         )
 
-    def get_frame_width(self):
-        return int(self._opencv_video_capture_object.get(3))
+    @property
+    def image_width(self):
+        try:
```
endurance, line 153: if we try to grab the image width and it fails, theres an issue with the software and we should raise an exception from the class.

src/cameras/detection/cam_detection.py
```diff
@@ -15,27 +15,24 @@ def find_available_cameras(self) -> FoundCamerasResponse:
...
         return FoundCamerasResponse(
-            camera_found_count=len(cams_to_use_list),
-            cams_to_use=cams_to_use_list,
+            number_of_cameras_found=len(cams_to_use_list),
+            list_of_usb_port_numbers_with_cameras_attached=cams_to_use_list,
             cv2_backend=cv2_backend,
         )
 
-    def _determine_backend(self):
+    @staticmethod
```
endurance, line 34: I dont think notating this as a static method helps since the method is private

src/cameras/detection/models.py
```diff
@@ -3,11 +3,7 @@
 from pydantic import BaseModel
 
 
-class RawCamera(BaseModel):
```
endurance, line 6: why get rid of RawCamera?

src/core_processor/camera_calibration/calibration_diagnostics_visualizer.py
```diff
@@ -0,0 +1,152 @@
...
+from pyqtgraph.dockarea.Dock import Dock
+from pyqtgraph.dockarea.DockArea import DockArea
+from pyqtgraph.Qt import QtWidgets
+
+import cv2
+
+from src.core_processor.camera_calibration.calibration_dataclasses import LensDistortionCalibrationData
+
+
+class CalibrationDiagnosticsVisualizer:
```
endurance, line 13: i have a lot of questions about this file, but ill come back to it later

src/core_processor/camera_calibration/camera_calibrator.py
```diff
@@ -0,0 +1,24 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class CameraCalibrator:
+    def __init__(self):
+        self.lens_distortion_calibrator = LensDistortionCalibrator()
+
+    def calibrate(self, this_camera: OpenCVCamera):
+        if self.lens_distortion_calibrator is None:
```
endurance, line 18: why would this ever be `None`? We definitely want to avoid "setup" kinds of workflows within "work" workflows to keep the code simple.

src/core_processor/camera_calibration/charuco_board_detection/charuco_board_definition.py
```diff
@@ -0,0 +1,10 @@
+import cv2
+
+aruco_marker_dict = cv2.aruco.Dictionary_get(cv2.aruco.DICT_4X4_250)
```
endurance, line 3: this would be a wonderful class.

src/core_processor/camera_calibration/charuco_board_detection/charuco_board_definition.py
```diff
@@ -0,0 +1,10 @@
...
+number_of_squares_width = 7
+number_of_squares_height = 5
+black_square_side_length = 1
+aruco_marker_length_proportional = .8
+
+charuco_board_object = cv2.aruco.CharucoBoard_create(number_of_squares_width, number_of_squares_height, black_square_side_length, aruco_marker_length_proportional, aruco_marker_dict)
+number_of_charuco_corners = (number_of_squares_width - 1) * (number_of_squares_height - 1)
```
endurance, line 10: this would be a wonderful class.

src/core_processor/camera_calibration/charuco_board_detection/charuco_dataclasses.py
```diff
@@ -0,0 +1,25 @@
...
+    any_markers_found: bool = False
+    image_width: int = None
+    image_height: int = None
+
+
+@dataclass
+class CharucoFramePayload:
+    raw_frame_payload: FramePayload = FramePayload()
+    annotated_image: np.ndarray = None
+    charuco_view_data: CharucoViewData = CharucoViewData()
```
endurance, line 25: We dont need to assign data in the base case. The caller will fill out these objects when its time to construct them.

src/core_processor/session/session_manager.py
```diff
@@ -0,0 +1,137 @@
...
+        with self._open_cv_camera_manager.start_capture_session_single_cam(
+                webcam_id=webcam_id,
+        ) as session_obj:
+            fps_manager = FPSCamCounter(self._open_cv_camera_manager.available_webcam_ids)
+            fps_manager.start_all()
+            cv_cam = session_obj.cv_cam
+            writer = session_obj.writer
+            try:
+                while True:
+                    if not cv_cam.is_capturing_frames:
```
endurance, line 38: this could be a timing issue - we could call start within the cv cam, but not be capturing frames yet for a camera and the entire method would exit. i think this should be a `continue` with a debug log

src/core_processor/camera_calibration/calibration_diagnostics_visualizer.py
endurance: okay, coming back to this file:

I noticed that every single method is public and the init is rather large. Could you break apart the init into neat named functions?

freemocap-ui/src/layout/Navigator.tsx
jonmatthis: makes sense, will do. 

I've been using PyCharm's "refactor" functions for all the renaming. I'll keep an eye out for front-end changes in the future. 

If there's some easy way to configure things so that happens automatically, let me know

src/cameras/capture/opencv_camera/frame_grabber.py
jonmatthis: works for me!

src/cameras/capture/opencv_camera/opencv_camera.py
jonmatthis: (btw, this bit is roughly speaking the boundary between "pure" computer vision and robotics/"perception engineering")

jonmatthis: This convention also has the nice feature of playing nicely with "single camera" setups - If there is only one camera attached to the computer, the world coordinates are defined to be the same as the camera coordinates things should go smoothly enough from there

jonmatthis: Once we're getting estimates of  3d skeletons, we can use those to estimate the *actual* world by making some pretty squishy assumptions to figure out the location and orientation of the groundplane, so that we can align things with gravity/'the inertial reference frame'

src/cameras/capture/opencv_camera/opencv_camera.py
jonmatthis: agreed - this'll get sorted out when I refactor to align with that pipeline diagram I sent you earlier. I'll remove this stuff for now and make sure it leaves things functional

src/cameras/capture/opencv_camera/frame_grabber.py
jonmatthis: my only concern is that this method makes this a little less explicit because you need to know (or figure out) that the `frame` object contains `success` `image` and `timestamp`.

... but I see why that's a good tradeoff, because it feel sloppy to create new variable if we don't know for certain we're going to use them

src/cameras/capture/opencv_camera/opencv_camera.py
jonmatthis: I added a `OpenCVCameraError` exception to that file, but I've never worked with custom exceptions and I don't know how to evoke that error intentionally so I'm not totally certain it works


(I also don't know how to reference specific lines and/or commits here...)

src/cameras/detection/cam_detection.py
jonmatthis: makes sense, I'll remove that. 

PyCharm told me that it was a "static method" and I got excited 😅

src/cameras/detection/models.py
jonmatthis: see below

src/cameras/detection/models.py
```diff
@@ -3,11 +3,7 @@
...
-    webcam_id: str
-
-
 class FoundCamerasResponse(BaseModel):
-    camera_found_count: int
-    cams_to_use: List[RawCamera]
+    number_of_cameras_found: int
+    list_of_usb_port_numbers_with_cameras_attached: List[int]
```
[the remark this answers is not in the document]
jonmatthis: I forgot to maintain compatibility with the option to use videos or IP cameras, so I wasn't sure what work `RawCamera` was doing. I'll put it back

src/core_processor/camera_calibration/calibration_diagnostics_visualizer.py
jonmatthis: yes, definitely. 

This QT GUI thing is going to get moved out into the domain of the `session_manager` in the next refactor.

The giant `__init__` is a result of the fact that I didn't know what I was doing when I started, but now that I've figured out  how `pyqtgraph` works I'll be chopping the whole thing apart and basically rebuilding from scratch

src/core_processor/camera_calibration/camera_calibrator.py
jonmatthis: fixed

src/core_processor/camera_calibration/charuco_board_detection/charuco_board_definition.py
jonmatthis: indeed it would! 

...it'd be even more wonderful if I could figure out how to make it `frozen`, but I don't know how to do that while also generating the `charuco_board_object` on the basis of the class properties.

src/cameras/capture/opencv_camera/frame_grabber.py
endurance: if we've named a class `FramePayload`, it /should/ make sense that the instance of such a class may be named `frame` - thats the logic i'm employing atm.

src/cameras/capture/opencv_camera/opencv_camera.py
endurance: ah okay understood

src/cameras/capture/opencv_camera/opencv_camera.py
endurance: Custom exception creation isn't particularly necessary. If we were releasing an SDK, then I would worry about something like that, and that would be created nearer to the customer (a.k.a developer) API rather than in low level classes that we use to handle our shit :D

Mainly, if we try to call `.get` on `None` or an OpenCV object that isn't ready, an exception will naturally be thrown, and we want that. `image_width` should just perform the call to `.get` and leave try/except out.

src/cameras/detection/cam_detection.py
endurance: I see. Technically, since there is no reference to any `self` instance variables, it COULD be defined as static. thats the heuristic that Pycharm uses to recommend. Not always perfect :D

src/core_processor/camera_calibration/calibration_diagnostics_visualizer.py
endurance: you dont have to break it apart in this PR
if there are plans to do that soon, im good.

it was big so i was like ahhhh *screams*

src/core_processor/camera_calibration/charuco_board_detection/charuco_board_definition.py
endurance: you can make another class as a dataclass with fields that are frozen. Does that hit what you're looking for?

src/cameras/capture/opencv_camera/opencv_camera.py
jonmatthis: got it, thanks!

src/core_processor/camera_calibration/charuco_board_detection/charuco_board_definition.py
jonmatthis: oh yeah, totally! Thanks!

PR #137, opened 2022-05-12
adding functionality to create a folder of mediapipe annotated videos

aaroncherian:


freemocap/fmc_mediapipe_annotation.py
```diff
@@ -0,0 +1,99 @@
...
+                break
+
+            mediapipe_results = holistic.process(raw_image)
+            annotated_image = annotate_image_with_mediapipe_data(raw_image, mediapipe_results)
+
+            writer.write(annotated_image)
+
+    writer.release()
+    cap.release()
+
```
jonmatthis, line 98: I've started trying to get in the habit of adding an example of how to use these functions when writing code like this. 

i.e. via  _ 

```python
if __name__ == "__main__":
    example_of_how_to_use_the_functions_in_this_script.run()
    
```

Obviously that's tricky to do in our current method, because there's a lot of machinery to generate the `session` object that's hard to reproduce here 

(but mostly I just wrote this because I  wanted to see what it was like to put a comment on your code :P )

freemocap/fmc_runme.py
```diff
@@ -201,6 +204,10 @@ def RunMe(sessionID=None,
             else:
                 print('`runMediaPipe` set to False, so we\'re loading MediaPipe data from npy file')
                 sesh.mediaPipeData_nCams_nFrames_nImgPts_XYC = np.load(sesh.dataArrayPath/'mediaPipeData_2d.npy', allow_pickle=True)
+            
+            if save_annotated_videos:
+                fmc_mediapipe_annotation.annotate_session_videos_with_mediapipe(sesh)
```
jonmatthis, line 209: nice and clean 👍

freemocap/fmc_runme.py
```diff
@@ -279,6 +287,8 @@ def RunMe(sessionID=None,
 
     # %% Stage 5 - Use Blender to create output data files
     if stage <=5:
+
+
```
jonmatthis, line 291: this is what happens when I leave a comment and click the "start a review" button

jonmatthis (approved):
looks good to me!

PR #138, opened 2022-05-12
Aaron/origin alignment 

aaroncherian:
Code for aligning stuff with the origin

The debug option for the origin alignment (those four skeleton plots I showed in lab meeting) is set to false by default to make sure it doesn't interrupt the run, just an fyi


freemocap/fmc_origin_alignment.py
```diff
@@ -0,0 +1,283 @@
+
+import numpy as np
+import pandas as pd
+import matplotlib.pyplot as plt
+from pathlib import Path 
+from rich.progress import track 
+
+def align_skeleton_with_origin(session, mediapipe_skeleton_data_XYZ, good_clean_frame_number):
```
jonmatthis, line 8: This'll probably want to be a `class` at some point

freemocap/fmc_origin_alignment.py
```diff
@@ -0,0 +1,283 @@
...
+
+    #skeleton_data_path = session.dataArrayPath/'mediaPipeSkel_3d_smoothed.npy'
+    debug = True
+    right_heel_index = 30
+    right_toe_index = 32
+    left_heel_index = 29
+    left_toe_index = 31
```
jonmatthis, line 15: and we'll pull these numbers from some `MediapipeDefinitions` dataclass

freemocap/fmc_origin_alignment.py
```diff
@@ -0,0 +1,283 @@
...
+        """Put two vectors in, make a normal vector"""
+        normal_vector = np.cross(vector1,vector2)
+        return normal_vector
+
+    def create_unit_vector(vector): 
+        """Take in a vector, make it a unit vector"""
+        unit_vector = vector/np.linalg.norm(vector)
+        return unit_vector
+
+    def calculate_skewed_symmetric_cross_product(cross_product_vector):
```
jonmatthis, line 39: is this the part you got from stack overflow?

jonmatthis: in general, code like this that encapsulates a lot of logic in a single call is hard to debug. Better to caluculate each row independently and then stack them to get the final output

freemocap/fmc_origin_alignment.py
```diff
@@ -0,0 +1,283 @@
...
+        skew_symmetric_cross_product = np.array([[0, -cross_product_vector[2], cross_product_vector[1]],
+                                                [cross_product_vector[2], 0, -cross_product_vector[0]],
+                                                [-cross_product_vector[1], cross_product_vector[0], 0]])
+        return skew_symmetric_cross_product
+
+    def calculate_rotation_matrix(vector1,vector2):
+        """Put in two vectors to calculate the rotation matrix between those two vectors"""
+        #based on the code found here: https://math.stackexchange.com/questions/180418/calculate-rotation-matrix-to-align-vector-a-to-vector-b-in-3d"""
```
jonmatthis, line 47: ah, gotcha 👍

jonmatthis (approved):
I don't have time to dig in to all the nitty gritty of the math, but in general the code looks nice and clean! Super cool to see you developing some fluency working with the data on this level :D

freemocap/fmc_origin_alignment.py
aaroncherian: image

what I got from stack exchange was this equation for rotating some unit vector a onto unit vector b

aaroncherian: https://math.stackexchange.com/questions/180418/calculate-rotation-matrix-to-align-vector-a-to-vector-b-in-3d

jonmatthis: yeah, right on. It does seem to do the job, so it's good enough for now :)

PR #143, opened 2022-05-30
trimming video with ffmpeg for faster speed

xyqyear:
For me, trimming takes more time than the later steps combined. I tested this to be able to reduce 90% of trimming time.

### DIFF
--- a/freemocap/calibrate.py
+++ b/freemocap/calibrate.py
@@ -2,6 +2,7 @@ import os
 import copy
 import json
 import pickle
+import subprocess
 import glob 
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


xyqyear:
This requires you have ffmpeg binary in your path so it's not plug and play. Maybe we can prompt the user to find the ffmpeg binary like how we did with blender.

endurance:
Thank you for your contribution!

We've removed 90% of the old code so this branch no longer makes sense to merge. Though, the concepts discussed here will be used in the new system!

As such, we'll be closing this pull request. Please keep checking back for more updates as we're adding new Issue templates, Contribution Guidelines, and a Roadmap for the open source community to keep contributing to the `freemocap` project.

PR #157, opened 2022-06-28
Good frame and origin alignment

aaroncherian:


freemocap/fmc_good_frame_finder.py
```diff
@@ -0,0 +1,103 @@
+
+import numpy as np
+import matplotlib.pyplot as plt
+
+
+
+
+def find_velocity_values_within_limit(skeleton_velocity_data, velocity_limit):
```
jonmatthis, line 8: terminology-wise, this would better labeled as a "threshold" rather than a "limit."  

not a big deal, just pointing out

freemocap/fmc_good_frame_finder.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+    This function takes in a skeleton velocity data array and a limit and returns the indices of the values that are within the limit
```
jonmatthis, line 10: again terminology-wise, this would be "..below that threshold"

freemocap/fmc_good_frame_finder.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+    indices = []
+    for i in range(len(skeleton_velocity_data)):
```
jonmatthis, line 13: At some point, remind me to show you how to do this with logical indexing

this method totally works, but it's a tad bit slower than it needs to be (not by much, but probably noticable with bigger larger number)

freemocap/fmc_good_frame_finder.py
```diff
@@ -0,0 +1,103 @@
...
+        if abs(skeleton_velocity_data[i]) <= velocity_limit:
+            indices.append(i+1) #add 1 to account for the difference in indices between the position and velocity data
+    return indices
+
+def find_matching_indices_in_lists(list_1,list_2,list_3,list_4):
```
jonmatthis, line 18: I think this is probably going to run pretty slow.  again good enough for now, but we can chat about it for future iterations

freemocap/fmc_runme.py
```diff
@@ -249,8 +251,11 @@ def RunMe(sessionID=None,
             if place_skeleton_on_origin:
                 sesh.mediaPipeSkel_fr_mar_xyz_smoothed_unrotated = sesh.mediaPipeSkel_fr_mar_xyz.copy()
                 np.save(sesh.dataArrayPath/'mediaPipeSkel_3d_smoothed_unrotated.npy', sesh.mediaPipeSkel_fr_mar_xyz_smoothed_unrotated) #save data to npy
+                good_frame = fmc_good_frame_finder.find_good_frame(sesh.mediaPipeSkel_fr_mar_xyz, .6, debug = False) #.6 is an initial guess for the velocity. seems to be a safe bet for for all the cases I've tried
```
jonmatthis, line 254: again, just pointing it out for posterity's sake - We're introducing a conflict with the term "good frame" here, as that concept is now used in multiple ways within our code (here for the origin alignment, but there's another "good frame" referenced in the Blender code.

freemocap/fmc_skeleton_data_holder.py
```diff
@@ -0,0 +1,100 @@
+
```
jonmatthis, line 1: Is this related to the `origin alignment` code? 

If not, it should technically be in a different pull request. 

(obviously we haven't been that careful about keeping our commits and PRs separated by 'feature' like that, but we'll want to start thinking in that was as we move forward with this project)

jonmatthis (approved):
Looks good! Left some notes on naming and whatnot, but nothing substantive


...I don't know how to test this without merging it into main, so I'll just approve the merge and test it from there lol

freemocap/fmc_skeleton_data_holder.py
aaroncherian: yes, it's used in the origin alignment.

PR #215, opened 2022-08-03
Freemocap_wiki_Draft_1

kileyhartigan:
Here's what I made for a lab wiki last fall.


jonmatthis:
This is good reference, but I think it's a bit too far behind the main repo to merge. We can use your fork as a reference as necessary moving forward

PR #217, opened 2022-08-06
Jon/docs/basic mkdocs layout

jonmatthis:
Setting up basic mkdocs layout...

...however, we might want to move our documentation to another repo to keep this one lean? I want to have a lot of images and whatnot in our docs. 

...but if we *do* then does that make it harder to auto-generate docs from the docstrings we make in this repo? 

... I guess another option is to host the images and whatnot elsewhere and link via URL's, but I'm not sure of how to do that? 

...ANYWAY - install requirements and run `mkdocs serve` to see sample docs site :D


jonmatthis:
I think I'm going to move this to another repo, something like - `github.com/freemocap/freemocap_documentation`

jonmatthis:
We set this up out of  - https://github.com/freemocap/documentation 

it's hosted at https://freemocap.readthedocs.io

eventually want to get it over to https://docs.freemocap.org 

but FYI - there's not much actual content up there right now (2022-08-22) so don't get excited lol

PR #226, opened 2022-08-25
Correct use_saved_calibration parameter

philipqueen:
OLD_README.md currently refers to the parameter `use_previous_calibration`, but the correct parameter name is `use_saved_calibration`. Using the wrong parameter name gives the error `TypeError: RunMe() got an unexpected keyword argument 'use_previous_calibration'`. This PR changes the existing parameter reference to the correct word, and adds the parameter to the default parameter list. 

There have been multiple Discord questions around this, so even though this PR will be outdated soon due to the Alpha release, it may save some time while pre-alpha is still in use.

### DIFF
--- a/OLD_README.md
+++ b/OLD_README.md
@@ -185,7 +185,8 @@ def RunMe(sessionID=None,
         resetBlenderExe = False,
       	get_synced_unix_timestamps = True,
         good_clean_frame_number = 0,
-        bundle_adjust_3d_points=False
+        bundle_adjust_3d_points=False,
+	use_saved_calibration = False,
         ):
 ```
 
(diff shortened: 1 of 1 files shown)


OLD_README.md
```diff
@@ -185,7 +185,8 @@ def RunMe(sessionID=None,
         resetBlenderExe = False,
       	get_synced_unix_timestamps = True,
         good_clean_frame_number = 0,
-        bundle_adjust_3d_points=False
+        bundle_adjust_3d_points=False,
+	use_saved_calibration = False,
```
jonmatthis, line 189: is this spacing wrong? The preview looks right, but `line 189` looks like its missing 2 or 3 spaces

jonmatthis (changes requested):

OLD_README.md
jonmatthis: on closer inspection, i think the file is correct, there's just something screwy with the way it's showing in this PR?

jonmatthis (approved):

PR #249, opened 2022-09-29
Jon/post process data

jonmatthis:
create start of a post-processing pipeline for 3d skeleton data, including: 

- Gap filling
- butterworth filtering
- align skeleton to ground plane


jonmatthis:
Hey @aaroncherian - Could you take a look at this PR and think about how it might match up with the automated `.trc` maker that (I think) you made a while back?

PR #301, opened 2023-02-10
Fix basic run for mac

philipqueen:
Currently, two things block the full refactor from running on mac: 1) Mediapipe will not download on apple silicon and 2) the `ctypes.windll` call in `__main__.py` will not run on mac.
This PR fixes (1) by adding a conditional dependency to `requirements.txt` which installs 'mediapipe-silicone` for arm64 macs. It fixes (2) by checking that the system is windows before calling `ctypes.windll`, which is only used to set the taskbar icon.


philipqueen:
This is now very outdated due to the merge to main. Issue 2 was fixed during the merge, but issue 1 with mediapipe silicon still needs to be fixed. I'm closing this for now and will readdress issue 2 at a later date.

PR #318, opened 2023-02-24
single cammy complete!

trentwirth:
completed the first implementation of `single camera mode`

there are some scaling issues, but it works!


trentwirth:
I need to add more features to this before it's merged

PR #323, opened 2023-03-01
Automate existing tests with sample data

philipqueen:
Runs the existing tests in the freemocap/tests folder with sample data. The sample data has already been run through freemocap - the next step will be to process sample data during testing, and run tests on what is produced.

Right now, the tests cover the 2d mediapipe, 2d mediapipe, and total body COM data shapes, and verifies that all of the videos in the synchronized video folder are the same length.


philipqueen:
I fixed the path finding functions in the last commit, so that it now uses the existing path/name storage methods. It also uses the naming conventions from the latest updates to trent/singlecammy. 

The name changing broke the tests on the sample data because the sample data is later than the newest name changes. We will need to record new sample data with the updated naming conventions for the testing to function again.

philipqueen:
This is outdated with the new release. I've ported the relevant changes to a new branch based on the current software state. It still needs some fixes so I'm not going to PR yet, but I'm going to close this one and develop on the new branch

Issue #330, opened 2023-03-08
Implement YOLO/RoboFlow based cropping/ pre-processing step for 2d skeleton tracking


jonmatthis:
posssible entry point (where image is sent to `mediapipe`) is here : https://github.dev/freemocap/freemocap/blob/87e2d2796a42d5c747cc63daa64c7c013314ddad/freemocap/core_processes/detecting_things_in_2d_images/mediapipe_stuff/mediapipe_skeleton_detector.py#L121

Here's Chris's previous implementation: https://github.com/CicaloC/roboflow_mediapipe/blob/main/roboflow_mediapipe.py

@ksm2264

jonmatthis:
and you should just make a branch off of `jon/full-refactor` branch (or `main` if you check and that branch doesn't exist anymore)

jonmatthis:
Closed by PR #333

PR #333, opened 2023-03-13
Jon/try yolo

jonmatthis:
PR based on this Issue #330


jonmatthis:
@ksm2264  - Made a PR for the YOLO stuff :)

jonmatthis:
I added a basic `YoloTracker` class at - `freemocap.core_processes.detecting_things_in_2d_images.yolo_tracker.YoloTracker`

It is accessed by way of a `bool` keyword argument to the `mediapipe_skeleton_detector.process_folder_full_of_videos()` method here: 
`freemocap/core_processes/detecting_things_in_2d_images/mediapipe_stuff/mediapipe_skeleton_detector.py:135`

...which is fed down to the `...detect_skeleton_in_image()` method and the actual work is done here: `freemocap/core_processes/detecting_things_in_2d_images/mediapipe_stuff/mediapipe_skeleton_detector.py:117`



----

The basic person detection part works, but I'm not translating the `mediapipe` results properly so the data is borked. I also suspect that we need to add some buffer and possibly make the crop into a square for better results. 

Try running the GUI and use the `Import Videos` functionality (`ctrl-i`) to load in a video with a person in it - verify that the videos were loaded by checking the 'Directory View' and 'Active Recording Info' tabs,  Then click `Control Panel >> Process MoCap Data >> Process Motion Capture Videos` (watch your terminal window for output)

It will fail when it gets to the 'calibration' stage, but it will also generate an `annotated_videos` folder next to the `synchronized_videos` folder in the active recording directory showing output from the `YOLO-mediapipe` pipeline as implemented.

PR #337, opened 2023-03-14
Add YOLO based pre-cropping before skeleton tracking

ksm2264:
crop using yolo, 
run mediapipe on cropped image,
modify mediapipe results object using crop image info, 
create annotated image using modified results


jonmatthis:
Added this functionalitiy via `freemocap/skellytracker` :)

PR #345, opened 2023-03-20
Fix blender executable path input on mac

philipqueen:
A hot fix allowing the path to an executable to be found by the file dialog widget, fixing the issue brought up on discord [here](https://discord.com/channels/760487252379041812/760489602917466133/1087385656134729758), where the blender executable is not directly accessible on macs.

If the selected file pathstring contains "Blender.app", as it does on mac, the widget adds `/Contents/MacOS/Blender` to the end of the pathstring. This points to the correct location of the executable inside the app, which is usually hidden from mac users and unlikely to be different between users/versions of blender.


jonmatthis:
@trentwirth - could you check if this works on Linux? And if not, update @philipqueen's fix so it does?

trentwirth (commented):
looks good! Reviewed by @jonmatthis @trentwirth and @philipqueen

jonmatthis (approved):

PR #346, opened 2023-03-20
Dark mode compatibility

philipqueen:
This is a very basic version of dark mode compatibility. It uses the darkdetect package to do a cross platform check for dark mode, and if dark mode is selected, it uses a different style sheet. If dark mode is not detected, it uses the standard style sheet.

The only change between the two style sheets currently is that the text in input boxes shows up in white in order to be visible against the dark mode background


philipqueen:
This PR isn't a great solution because of the dependency on darkdetect, and because it will all be changed by the progress in trent/color-restyling. I'm going to scrap this branch and PR in favor of finding a solution in the scss method

PR #347, opened 2023-03-20
Trent/color restyling

trentwirth:
- changes to GUI recoloring
- improvements of font size 
- removal of unnecessary emojis

I recognize that there is more room to improve, but this GUI looks closer to what I imagine a professional product will look like. I relied more on the core colors of freemocap (as picked by Taylor the First) to guide button styling. What the "final" colors should be is for sure up for debate, but I think this GUI is better to look at because it's not as overwhelming as the previous version. I think it's a good candidate for the aesthetic of `1.0.0`

NOTE: There are two UI improvements that need to happen in *skelly cam* that I didn't implement here - 1. the emojis stripped out and 2. the buttons like `Expand All` and `Collapse All` are buttons on top of text, and it looks bad. So the background text needs to be removed (?)

image
image
image


trentwirth:
color updates: 

image
image
image

trentwirth:
image

trentwirth (commented):
looks great!! reviewed with @jonmatthis , @trentwirth  and @philipqueen

jonmatthis (approved):

PR #359, opened 2023-04-04
Add help and support menus

philipqueen:
This pull request adds help and support tabs back to the menubar. It uses the old commented out code, but updates it to match the style of the currently-in-use file menu. 

I was not able to find any way to have the support option be a button instead of a drop down ([this solution doesn't work](https://stackoverflow.com/questions/52986172/pyqt-about-button-in-menu-bar)).
Also, because of how Macs handle the menubar, an option that starts with the text "About" will not show up (it should show on windows). We can either change the text ("Freemocap - About Us", for example) or be content with it not showing up on mac.

Screen Shot 2023-04-04 at 12 10 51 PM
Screen Shot 2023-04-04 at 12 10 57 PM
Screen Shot 2023-04-04 at 12 11 03 PM

### DIFF
--- a/freemocap/gui/qt/actions_and_menus/actions.py
+++ b/freemocap/gui/qt/actions_and_menus/actions.py
@@ -1,4 +1,5 @@
-from PyQt6.QtGui import QAction
+from PyQt6.QtGui import QAction, QDesktopServices
+from PyQt6.QtCore import QUrl
 
 
 CREATE_NEW_RECORDING_ACTION_NAME = "New Recording"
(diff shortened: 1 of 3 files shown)


freemocap/gui/qt/actions_and_menus/actions.py
```diff
@@ -9,6 +10,11 @@
 REBOOT_GUI_ACTION_NAME = "Reboot GUI"
 EXIT_ACTION_NAME = "Exit"
 
+DOCUMENTATION_ACTION_NAME = "Open Documentation"
+ABOUT_US_ACTION_NAME = "About Us"
+
+DONATE_ACTION_NAME = "Donate to Freemocap"
```
jonmatthis, line 16: Make sure these action names match their connected part of the Menu (not saying they don't now because I didn't check, just saying to make sure they match :)  )

jonmatthis (changes requested):

freemocap/gui/qt/actions_and_menus/actions.py
philipqueen: They do now!

philipqueen:
Changed the "About Us" to "Freemocap Foundation" to give it consistent behavior on Macs

Screen Shot 2023-04-05 at 10 17 23 AM

jonmatthis (approved):

PR #393, opened 2023-05-08
adding jupyter notebook generation

### LINKED ISSUE #390: Autogenerate Jupyter notebook for each recording
closes #393

jonmatthis:
This PR is to add a feature that automatically generates a jupyter notebook for each recording by pulling from a template `ipynb` somwhere, updating the paths to the data, and then plopping into the recording folder. 

This will probably be added as an `export` option (alongside the 'blender' output) 

In the codebase, we'll want to put it in the folder `freemocap/core_processes/jupyter_stuff`

in the GUI, this will want to live under 'Export" tab,  I think. It will either be a checkbox for now, but we can add a parameter tree later if it picks up any parameters. 

closes #390 

Tasks:

- [x] Create a Class or function (your call) that takes in a recording folder path, generates a basic juptyer notebook to load the data from that path (assuming standard paths for now)  and saves it to the recording folder (it should run from an `if name equal main` block with a default folder path 

- [x] Incorporate method into standard processing pipeline

- [x] Add a checkbox or something to the GUI that lets people turn it off if they want

- [x] Create a nice standard jupyter output that:
    - [x] that makes plotly plots of joint trajectory (example of how to view a timeseries with `plotly`    
    - [x] Plays some kind of interactive 3d viewer of skeleton data (e.g. with `ipywidgets` or something? I think @trentwirth  did that at some point?) (example of how to view 3d data with `matplotlib`)
    - [ ] BONUS - also plays some/all annotated  videos in the notebook (but that might not be technically feasible? Hard to know)(example of way to load a view a video)


jonmatthis:
closes #393

aaroncherian:
@jonmatthis Tested this with a single camera recording (both having `generate jupyter notebook` checked and unchecked), and it completed successfully.

jonmatthis:
merging this onto my VSS branch for now - I want to learn about more about how 'papermill' works before merging into main

PR #396, opened 2023-05-09
Fix invalid escape sequences

philipqueen:
The end to end testing is giving a warning about two invalid escape sequences: `DeprecationWarning: invalid escape sequence`. There were two backslashes in strings that were not followed by a valid escape sequence. One is in a path that needed a raw escape, and the other was a backslash that appeared to be a newline missing the n, which I added.

This is pretty minor code cleanup to appease the test suite, but someone on windows should test the blender path finding still works properly.


philipqueen:
This was put directly into the end to end testing PR by Jon, so this can be closed!

PR #400, opened 2023-05-16
Material colors styling

philipqueen:
Working towards a more modern, Material styled GUI.

This includes:
- reserving the use of colors to draw the eye to key elements
- hierarchical button styles
- button's react in meaningful ways when hovered, selected, or pressed

Screen Shot 2023-05-16 at 4 29 17 PM

### DIFF
--- a/freemocap/gui/qt/style_sheet/qt_style_sheet.css
+++ b/freemocap/gui/qt/style_sheet/qt_style_sheet.css
@@ -35,7 +35,8 @@ QTabBar::tab {
   font-weight: bolder;
   font-size: 16px; }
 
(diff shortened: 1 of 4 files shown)


jonmatthis:
@philipqueen - this really looks great! 

I'd like to have this merged before I go to Florida (Thurs evening), so go ahead and do whatever last fiddly bits you want to it before then. It doesn't have to be exhaustive, we can (and likely will) always keep working on it after

jonmatthis (approved):
Look great!

I think we'll want to make a "Dark mode" and other color styling options at some point, for now this is a really nice simple sleek and 'modern looking' design ✨

PR #401, opened 2023-05-17
Automate end to end testing

### LINKED ISSUE #389: Automate testing with Github Actions

philipqueen:
This would close #389 by automating the newly added end to end testing. 

Right now, it is set to run whenever a PR is made to main that touches the `freemocap` folder. If that runs the tests too often and eats into our allotted github runner time we can think of a better way to determine when to run the tests.

### DIFF
--- a/.github/workflows/python-testing.yml
+++ b/.github/workflows/python-testing.yml
@@ -23,7 +23,7 @@ jobs:
           pip install -e .
       - name: Replace Opencv with Headless
         run: |
-          pip uninstall opencv-contrib-python
+          pip uninstall -y opencv-contrib-python
           pip install opencv-contrib-python-headless
       - name: Run Tests with Pytest
         run: |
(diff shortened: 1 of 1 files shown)


philipqueen:
This testing is now working, and takes about 2.5 minutes to run. 

It appears from here that because this is a public repository, we don't have a usage limit: https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions
Even with the lowest usage limit, we could still run tests 800 times a month though.

jonmatthis (approved):
Nice!

So this means that when we try to merge a PR, it'll run the test and then reject the merge if the tests fail?

philipqueen:
@jonmatthis not quite. When someone makes a PR it will run the tests, and it will continue to run tests as they commit to the PR. To make it so only passing PRs can merge, we can change the "Require status checks to pass before merging" branch protection rule, which will not allow merging PRs with a failing test suite

PR #402, opened 2023-05-18
Testing GitHub action

philipqueen:
This creates a dummy test file in the freemocap/tests folder to trigger the github action. The point of this is to make sure the action in philip/automate-end-to-end-testing #401 runs properly before merging it into main.


philipqueen:
Testing on the relevant branch is working, this is no longer needed

PR #404, opened 2023-05-23
Save parameter state

### LINKED ISSUE #387: Save previous state (i.e. user settings)

philipqueen:
This PR saves the recording processing parameter model to a json file at the end of processing. This has a few benefits:
1. It allows the user to check what parameters they used to process a session after the fact.
2. It will help users debug problems in the processing or output.
3. It will allow us to save the parameters a user choose to preload them when they reopen the app.
Would close #387 

Tasks:
- [x] Save the state of the parameter tree to JSON
- [x] OPTIONAL - move the parameter saving to the start of the processing pipeline: Right now its at the end with the other `save_dictionary_to_json` calls. But if processing fails, the JSON will not be saved so it will not help with debugging. Saving it at the start sounds better but I want feedback on this.
SEPARATE PR - Load the last saved parameter state: Right now it just saves the dictionary to the output data folder, but I'm working on copying the 'last successful calibration' model to also automatically load the last saved parameter state back into the parameter tree on loading the app. --- We're going to handle this in a new PR with the whole gui state being saved and then set at reload
SEPARATE PR - Save and load the camera state as well: I'm assuming this will require work on skellycam instead of freemocap core, so I'm reserving it for another PR once I have this one working.

Example JSON:
```json
{
    "mediapipe_parameters_model": {
        "model_complexity": 2,
        "min_detection_confidence": 0.5,
        "min_tracking_confidence": 0.5,
        "static_image_mode": false,
        "skip_2d_image_tracking": true
    },
    "anipose_triangulate_3d_parameters_model": {
        "confidence_threshold_cutoff": 0.5,
        "use_triangulate_ransac_method": false,
        "skip_3d_triangulation": true
    },
    "post_processing_parameters_model": {
        "framerate": 30.0,
        "max_gap_to_fill": 10,
        "skip_butterworth_filter": true,
        "butterworth_filter_parameters": {
            "sampling_rate": 30.0,
            "cutoff_frequency": 7.0,
            "order": 4
        }
    }
}
```


jonmatthis (approved):
looks good!

PR #408, opened 2023-05-31
Error catching in headless processing

philipqueen:
This fixes the TODO in `process_recording_headless.py` by raising an error if a multi-video recording is being processed without a provided calibration toml.

The code change itself is quite simple, but I can't test it on a single camera recording because of an error with papermill I'll create an issue for.


jonmatthis (approved):
Looks good to me!

I guess we could also set up a 'single-camera' end-to-end test too now, which would help increase our test coverage (by hitting the 'single camera' parts of our code that get skipped in our current e2e test)

PR #421, opened 2023-06-07
Save gui state

philipqueen:
This PR adds a `gui_state.json` file that stores the state of parameters in the gui, and when the gui reopens loads the last parameter state. The state is resaved whenever a parameter is changed, so it remains up to date even after a crash.

Some potential improvements include:
- [x] Adding a "return to default" option in the menu bar
- [ ] There are two charuco square size options now, and they could be linked so that changing one changes the other


philipqueen:
Add a menu option to reset defaults

philipqueen:
Screen Shot 2023-06-20 at 1 34 12 PM
Here is the menu option to reset to defaults

jonmatthis (approved):

PR #429, opened 2023-06-22
Test across versions with nox

philipqueen:
[Nox](https://nox.thea.codes/en/stable/index.html) is a tool that allows for automated testing across multiple python environments and different package versions. It is similar to tox, but uses python instead of an. `.ini` file to set up. 

The nox file in this PR currently runs the full test suite across python versions 3.8 through 3.11. To run it, first deactivate conda/whatever virtual environment manager you use (nox will create a virtualenv for each test, so you want it running globally). If you haven't used it before, you'll then want to install it with `pip install nox`. Then cd to freemocap (with this branch checked out) and simply run `nox`. 

To-Do List:
- [x] Setup nox to run our current test suite in each python versino that we say we support (currently 3.8+ according to the README)
- [ ] Setup nox in our github action test runner. This is easy, but we have to decide which tests we want to run where. Maybe we only run a full, across version check right before merging. Since nox runs the full tests in each environment, it is quite time intensive. We can setup the github runner to only run a subset of the full nox tests, or set it up to run different tests across different hooks.
- [ ] Decide if there's packages we want to support multiple versions of, for example skellycam. We can test across multiple versions of packages, as part of or separate from the testing across different python versions.
- [x] Add a coverage check session. We might as well have the pytest coverage run that we want to add run in Nox. This won't have to happen across multiple python versions, so it should be in a separate session.
- [ ]  (SEPARATE PR) Set up a lighter weight  test suite. We should have unit tests for our functions and classes, and be running these across a wide variety of python and package versions, and only run our end to end tests on a subset of these.
- [ ] (SEPARATE PR) Add a linting/formatting step to the nox run. Right now we don't want to lint with, say Black, because it will want to make an overwhelming amount of changes at once. But once most of our code conforms to the Black standard, we may want to lint new code.


jonmatthis (approved):
looks good

philipqueen:
The x on the final test is because I cancelled it to stop us from running a third round of 20 minute tests in a row, not because any tests failed

PR #432, opened 2023-06-26
improve save_dictionary_to_json to pathlib native syntax

newturok:


philipqueen:
Hi @newturok, thanks for contributing to Freemocap! Can you add some reasoning as to why you've made the change? The current method has the bonus of using the `with` context manager, which is the standard for file i/o in python. If you can share arguments for your change though we're happy to consider them.

newturok:
@philipqueen just because Path.write_text has some additional guards under the hood and is cleaner 
image

jonmatthis (approved):
Nice! I didn't realize that pathlib could do that - thank you!!

PR #437, opened 2023-06-28
adding postprocessing improvements

aaroncherian:


freemocap/core_processes/post_process_skeleton_data/calculate_center_of_mass.py
```diff
@@ -0,0 +1,383 @@
...
+    return mediapipe_frame_segment_joint_XYZ
+
+
+# %%
+# values for segment weight and segment mass percentages taken from Winter anthropometry tables
+# https://imgur.com/a/aD74j
+# Winter, D.A. (2005) Biomechanics and Motor Control of Human Movement. 3rd Edition, John Wiley & Sons, Inc., Hoboken.
+
+
+BODY_SEGMENT_NAMES  = [
```
jonmatthis, line 181: Disclaimer: I haven't looked at this part toooo deeply, but I had a sloppy thought, re: "calculating COM from arbitrary skeleton input" 

You could probably build a dictionary (or Pydantic model) that uses this list as the `keys` and something akin to the model in `freemocap.data_layer.data_saver.data_models.VirtualMarkerDefinition`  

(For reference, that's a Pydantic version of the `mediapipe_virtual_marker_definitions_dict` defined here as a good ol' fashioned dictionary:` freemocap/data_layer/export_data/blender_stuff/blender_bpy_export_scripts/freemocap_blender_megascript_take2.py:788`)

That way, you could have a concise schema (save and loadable as a `json`) that could define how to combine the `mediapipe` markers to get the joint_centers necessary to calcuate the COM - i.e. it would be a compact way to define the interface between 'mediapipe shaped data' and the dataframe defined in the `build_anthropometric_dataframe()` function (which can also be saved/loaded as a json or csv)

Think about whether that makes sense and we can chat about it later :)

jonmatthis: Oh, just realized - the above story might want you to make another list called `BODY_JOINT_CENTERS` or something

jonmatthis: ...which could similarly be a dictionary with values corresponding to your `joint_connections ` variable below

freemocap/core_processes/post_process_skeleton_data/calculate_center_of_mass.py
```diff
@@ -0,0 +1,383 @@
...
+        totalBodyCOM_frame_XYZ[frame, :] = this_frame_total_body_COM
+
+    f = 2
+    return totalBodyCOM_frame_XYZ
+
+
+# %%
+
+
+def build_anthropometric_dataframe(
```
jonmatthis, line 317: Can we save this dataframe out somewhere with a `to_csv()` or `to_json()` call? 


(see longer comment above for context)

...After looking at this again, I realize this function is more or less doing what I was requesting, that is it's generating a nice schema based on the anthropometic data defined as variables above and then using that to calculate the COM later (or something like that) 

I think what I'm proposing now is to just save this data out upon calculation - at first as a nice record of the schema used in the calculation, and later we might just have a folder of `json`s or `toml`s or something that includes various definitions of COM anthropometic definitions (e.g. from other papers, or adjust for prosthetics, etc) as well as schemas to define interfaces between a given anthropometric skeleton schema and a given skeleton data schema (e.g. mediapipe, qualisys, etc) 


Again - these are on the fly thoughts, I might be missing something

freemocap/core_processes/post_process_skeleton_data/calculate_center_of_mass.py
```diff
@@ -0,0 +1,383 @@
...
+    trunk_proximal = (trunk_XYZ_coordinates[0] + trunk_XYZ_coordinates[1]) / 2
+    trunk_distal = (trunk_XYZ_coordinates[2] + trunk_XYZ_coordinates[3]) / 2
+
+    return trunk_proximal, trunk_distal
+
+
+# %%
+
+
+def build_mediapipe_skeleton(mediapipe_pose_data, segment_dataframe, mediapipe_indices) -> list:
```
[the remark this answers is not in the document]
jonmatthis: No, definitely not - just thinking about the future as I'm going over these 💫

jonmatthis (changes requested):
I'm was about ready to approve your branch, but when I tested with a recording I made while sitting down, it *worked* but returned a bunch of wonky data for the legs and whatnot

I think that's also the source of all the `mean of an empty slice` warning that  @philip_q  mentioned above 


Questions; 
- Is this a new behavior that comes from `skellyforge`? Or did the old `post_processing` methods also have this issue? 

- If so, can you check for the (likely sloppy) method we used to fix it in the previous code and think about re-implementing it in either `skellyforge` or `freemocap`? (If it's a sloppy method, consider something new, like a reprojection_error filter (see below)


Possible causes/fixes

 - it's possible that mediapipe is still assigning my legs a confidence value greater than `0.5` even if it doesn't ever seem them. Could be fixed by making the medipaipe_confidence threshold more restrictive (i.e. put the default at `.7`  instead of `.5`, though I know that caused issues before...) 

- Could be fixed by implementing some "reprojection error filter" (either in skellyforge or freemocap) that replaces wackadoo 3d data points with `nans`.  In order to avoid the "over filtering" issue we ran into a while ago, we could could be made to be very permissive, say only filtering dottos that are greater than 5-10*the standard deviation of the rest of the projection error data

jonmatthis (approved):
Apparently that issue mentioned in the previous review has been a problem for a while so we can approve this PR and work on fixing that issue in the future!

aaroncherian:
> I'm was about ready to approve your branch, but when I tested with a recording I made while sitting down, it _worked_ but returned a bunch of wonky data for the legs and whatnot
> 
> I think that's also the source of all the `mean of an empty slice` warning that @philip_q mentioned above
> 
> Questions;
> 
> * Is this a new behavior that comes from `skellyforge`? Or did the old `post_processing` methods also have this issue?
> * If so, can you check for the (likely sloppy) method we used to fix it in the previous code and think about re-implementing it in either `skellyforge` or `freemocap`? (If it's a sloppy method, consider something new, like a reprojection_error filter (see below)
> 
> Possible causes/fixes
> 
> * it's possible that mediapipe is still assigning my legs a confidence value greater than `0.5` even if it doesn't ever seem them. Could be fixed by making the medipaipe_confidence threshold more restrictive (i.e. put the default at `.7`  instead of `.5`, though I know that caused issues before...)
> * Could be fixed by implementing some "reprojection error filter" (either in skellyforge or freemocap) that replaces wackadoo 3d data points with `nans`.  In order to avoid the "over filtering" issue we ran into a while ago, we could could be made to be very permissive, say only filtering dottos that are greater than 5-10*the standard deviation of the rest of the projection error data

So I think the main underlying issue here is one we had talked about awhile ago actually - which is that the interpolation/filtering was mainly built for our research purposes, where we assumed a full body view would be used. Essentially, the Butterworth Filter won't work if there are any NaN values - so the interpolation function does two things

1) it interpolates between the data it can see, and then 
2) replaces any NaN values left over with mean values of the array at that time (in our research recordings, this is usually at the start and end of the recording when the person is walking into/out of frame)

Both of these are needed to get through the butterworth filtering.

However, when you have a 'sitting at your desk' recording - where the legs are never seen, this causes the second step to go a little haywire in filling in all of those NaN values. This is also what causes the 'mean of an empty slice' error. 

So basically - the interpolation/filter functions were made to work on 'whole body' skeletons, since that's what we needed it for in research, and if we want it to work for something like a 'sitting at a desk' recording, we would need a different/smarter way about determining what joints get fed into interpolating/filtering.

PR #440, opened 2023-06-30
Pin dependency versions

philipqueen:
This PR pins the versions of mediapipe and libsass in our dependencies to the last working versions.

Mediapipe is make breaking changes to its models now. They haven't touched holistic yet, so this version works, but we can't guarantee how long the most recent mediapipe will work. We're freezing it at `0.10.1` to ensure it works, and we can update that as new working versions come in. We will eventually want to transfer to the new mediapipe solutions, but we'll do that as part of the skellytracker update. 

Also, mediapipe now includes arm64 support in its builds, so we're getting rid of the conditional mediapipe-silicon dependency as well.

Libsass `0.22.0` no longer runs on arm64, even though the prior version does, so we're pinning that to `0.21.0` until a more recent version works across the systems we want to support.


philipqueen:
I just added pydantic as an explicit dependency so I could specify it to v1*. This is to avoid problems with the v2 rollout, which makes breaking changes outlined here: https://docs.pydantic.dev/latest/migration/

aaroncherian:
I gave this a test run and was able to get the GUI running!

jonmatthis (approved):

PR #441, opened 2023-07-03
Change README to always install latest pip pre release

philipqueen:
This PR changes the `pip install freemocap~=.....` instructions with just `pip install --pre freemocap`, as this will install the latest dev branch until we make the 1.0 release. This way we can avoid having to change the install instructions with each pip release. I've verified `pip install --pre freemocap` downloads 1.0.14rc0 right now.


jonmatthis (approved):
looks good to me!

PR #447, opened 2023-07-06
Add notes about testing in CONTRIBUTING.md

philipqueen:
This PR adds a note about running our test suite locally before submitting PRs. Since our automated tests only run with organizational approval, this will allow contributors to know sooner if their contributions are working. We could also consider making a whole `testing` section where we also discuss running the Nox tests, but I've made the changes minimal for now.


jonmatthis (approved):
I want to double check the `pip install -e '.[dev]'`  before merging, but I think its fine

PR #453, opened 2023-07-14
Lint everything with Black

philipqueen:
"I see a red door and I want it painted black"

We've been roughly trying to follow black, but here's full compliance! We'll add testing for this in Nox soon, with the flake8 branch


jonmatthis (approved):
its probably fine lol

PR #460, opened 2023-07-27
Welcome screen

philipqueen:
Add a welcome screen to greet new users with links to the documentation. 
It has a "Don't show this again option" which is saved in the gui state, so it will appear any time the app is opened until that checkbox is clicked, and then will not appear. I tested that adding this new field to the gui state will not cause problems with the existing gui state problems that do not have this field - it will just be initialized with the new field default as it's loaded from file.

Screen Shot 2023-07-27 at 1 37 15 PM


philipqueen:
Screen Shot 2023-07-27 at 1 49 49 PM

I added a message about the sample data. We should have documentation about the sample data as well, that goes through how to load the data and process it, along with simple troubleshooting.

philipqueen:
Screen Shot 2023-07-28 at 12 11 16 PM
I added a little welcome message to the top, it can be expanded if we'd like

jonmatthis (approved):

PR #464, opened 2023-07-31
PATCH: fix batch processing error

### LINKED ISSUE #462: Batch processing expects an output data folder to exist
So I was trying to use the batch processing code, and I ran into this error:

`Exception has occurred: FileNotFoundError
[Errno 2] No such file or directory: 'D:\\2023-07-26_SBT002\\1.0_recordings\\calib_1\\sesh_2023-07-26_14_19_10_STB002_NIH_Trial1\\output_data\\recording_parameters.json'
  File "C:\Users\aaron\Documents\GitHub\freemocap\experimental\batch_process\batch_process.py", line 65, in process_recording_without_gui
    save_dictionary_to_json(
  File "C:\Users\aaron\Documents\GitHub\freemocap\experimental\batch_process\batch_process.py", line 40, in process_folder_of_session_folders
    process_recording_without_gui(
  File "C:\Users\aaron\Documents\GitHub\freemocap\experimental\batch_process\batch_process.py", line 92, in <module>
    process_folder_of_session_folders(
FileNotFoundError: [Errno 2] No such file or directory: 'D:\\2023-07-26_SBT002\\1.0_recordings\\calib_1\\sesh_2023-07-26_14_19_10_STB002_NIH_Trial1\\output_data\\recording_parameters.json'`

This happens in the `save_dictionary_to_json` function in `batch_process.py`. As far as I understand, the function is trying to save the `json` into the `output_data` folder, but that folder doesn't exist yet, so we get this error.

This file used to work fine, so my guess is that there's a recent update that caused this to stop working. I've done a manual workaround to create an `output_data` folder in each session, and that works fine, but we should still try and make a fix for it, even if its just adding a pathlib line to create the `output_data` folder.

philipqueen:
When saving the recording parameters, you need to make sure the output data folder is created before trying to save a file there. We made this change everywhere else when we added that feature, but must have missed this. 

In the future, should we consider just creating the output data folder when we create the session folder? This would prevent having to duplicate this code across so many different places. Alternatively, we could have the save to json function create the output folder if it doesn't exist yet. Both are DRYer than this method.

For now though, this fixes #462

### DIFF
--- a/experimental/batch_process/batch_process.py
+++ b/experimental/batch_process/batch_process.py
@@ -62,6 +62,8 @@ def process_recording_without_gui(
 
     recording_info_dict = rec.dict(exclude={'recording_info_model'})
 
+    Path(rec.recording_info_model.output_data_folder_path).mkdir(parents=True, exist_ok=True)
+
     save_dictionary_to_json(
         save_path=rec.recording_info_model.output_data_folder_path,
         file_name=RECORDING_PARAMETERS_JSON_FILE_NAME,


jonmatthis (approved):
👍 

In general (I think ) its good practice to run that 'mkdir' bit before saving any files to disk 

it protects from this kind of error, and also protects from creating empty folders (bc you don't don't create the folder until it's time to save something to it)

PR #466, opened 2023-08-03
Reprojection Error Filtering

philipqueen:
Updated filtering based on thresholding the mean reprojection error. This method finds frames that have a reprojection error above a given threshold (in linear units), and retriangulates those frames while removing data from one or more cameras until all frames are under the given threshold. It first goes through all of the combinations with 1 camera removed, then all the combinations with two cameras removed, and so on. It stops either when there are no longer reprojection errors below the threshold, or when it has exhausted all of the two camera combinations. If it has exhausted all two camera combinations, it sets remaining values above the threshold to NaNs, which allows them to be gap filled in post processing. Because it only retriangulates frames above the threshold on each iteration, it is fairly fast compared to the original triangulation.

Potential Improvements:
- [x] Better default value. Right now I've chosen 150mm as the default, as this value appears sensible on the sample data and my own go to test session. However, we should compare across more sessions before merging, and err on the side of less restrictive. - The default is now 90%, so 10% of the data is refiltered.
- [x] Allow filtering based on some deviation from the mean. This should be simple to implement, as we can just calculate the linear value based on the mean and pass that into the existing function. - Calculates the mm value for a given percentile
- [x] A mechanism for either choosing which cameras to prioritize moving, or a way of telling which cameras are contributing the most to the error and removing those. Right now we're only getting the mean across cameras of the reprojection error per marker from anipose, so we can't access the per camera reprojection error. - FUTURE PR

Debug plot from sample data:
debug_reprojection_error_filtering

Parameters in gui:
Screen Shot 2023-08-03 at 10 46 35 AM


philipqueen:
Just merged the by camera reprojection filtering into this

PR #471, opened 2023-08-25
Fix calibration toml finding

philipqueen:
This PR fixes the problem we've had with not being able to automatically find the calibration toml paths of an active recording session. If a path is not found at `SESSION_NAME_camera_calibration.toml`, it checks the active recording folder for another file that ends with `camera_calibration.toml`. If it finds one, it sets that to the calibration path for the active recording, and if not it defaults to the most recent calibration. 

It also fixes the linting that got pushed this morning, which is why it touches some files unrelated to the calibration path.


jonmatthis (approved):
Looks good to me!

(until we refactor this whole nonsense in later, anyway)

PR #472, opened 2023-09-04
add `feedback` and `donation` link, check for new version

jonmatthis:
- add `feedback` form (https://forms.gle/AguGxQAJGXxaJbXdA) and `donate` link to home tab 
- check for new version of freemocap

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,10 +1,12 @@
 [flake8]
-
+root = ./freemocap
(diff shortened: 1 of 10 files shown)


jonmatthis:
passes main test - fails flake8 (but passes on my machine after editing the .flake8 file to exclude venv)

PR #473, opened 2023-09-05
Update opencv version to 4.8

philipqueen:
We've had opencv pinned to version `4.6.*`, despite having `try/except` blocks that handle the namespace changes from versions `> 4.6.*`. This update changes the version to 4.8 to allow us to benefit from ongoing updates to the charuco functionality in opencv.

Unfortunately, we can not change the dependency from `opencv-contrib-python` to `opencv-python` as we had originally hoped, because both mediapipe and anipose rely on `opencv-contrib-python`, so it will only add an additional dependency.

### DIFF
--- a/freemocap/core_processes/capture_volume_calibration/anipose_camera_calibration/anipose_camera_calibrator.py
+++ b/freemocap/core_processes/capture_volume_calibration/anipose_camera_calibration/anipose_camera_calibrator.py
@@ -3,7 +3,6 @@ from pathlib import Path
 from typing import Callable, Union
 
 import numpy as np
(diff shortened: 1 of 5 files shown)


philipqueen:
I've removed the try/except blocks to handle `opencv=4.6.*`, last check is to make sure we can calibrate from webcams without issue.

philipqueen:
I've ported more aniposelib code into our `freemocap_anipose` function. The ones on lines `~200-350` are to fix the zsh bus error on macs. The changes from `~1900` on are to get the charuco functions working with opencv. Both are changes we can revert if aniposelib is open to changes in the future.

jonmatthis (approved):

PR #478, opened 2023-09-14
Annotate charuco board on images

### LINKED ISSUE #411: Annotate images with Charuco points

philipqueen:
Draws charuco points on images when checkbox is checked. This should allow users to diagnose charuco problems visually before recording. It's been added to the gui state as well, so the checkbox status will be preserved across application sessions. To refresh the annotation status with cameras running, you have to close and reopen the cameras.

Closes #411 

Screen Shot 2023-09-14 at 2 11 01 PM


philipqueen:
I was finally able to test this with webcams. I had no lag with three webcams with the annotation turned off, but noticeable lag (a few frames a second?) with the annotation turned on.

I got the charuco to display well with my built in webcam, but not with my cheaper webcams

PR #483, opened 2023-09-22
Update README.md

FangSen9000:
Quick start here is confusing, I thought you didn't need to clone code


philipqueen:
Thanks @FangSen9000, but you don't need to clone freemocap to pip install it. Doing both without keeping environments separate can cause problems as well. We'll keep the pip install instructions for quickstarting, and separate build from source instructions for cloning

PR #486, opened 2023-09-29
Calibration filtering by reprojection error

philipqueen:
Merge the by camera reprojection filtering into the frontend backend pr


philipqueen:
I've made most of the changes to get this working with the changes in the process recording folder, as well as some other catch up changes. The only remaining hurdle is a blender issue, likely to do with changing the path names. I'll look into it soon.

philipqueen:
Just tested blender again and it ran fine on the sample data and a single camera recording, so whatever error I saw last may have been unrelated. If someone else can test this and see if blender runs fine it should be good to merge into development

PR #488, opened 2023-09-29
Brightness synching into frontend/backend PR

philipqueen:
move brightness synching into frontend/backend PR


jonmatthis (approved):
Looks good to me!

PR #492, opened 2023-10-09
PATCH: offline failure

philipqueen:
This patches the issue brought up on Discord that the gui will not currently open if the user's computer is not connected to the internet. I've added exception catching for the version checking and pipedream pings to fail gracefully when not connected to the internet. I've also changed the error handling with the sample data downloading to fail gracefully there as well.


jonmatthis (approved):
looks good!

PR #503, opened 2023-11-25
Merge main history changes into development

philipqueen:
The git history changes need to be pulled into development so that we can keep working there. This should be a simple merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <p align="center">
     <img src="https://raw.githubusercontent.com/freemocap/freemocap/main/assets/logo/freemocap-logo-black-border.svg" height="64" alt="Project Logo">
-</p>
+</p> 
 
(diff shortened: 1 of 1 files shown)


philipqueen:
@jonmatthis the github action is failing on `[tool.poetry] section not found in /home/runner/work/freemocap/freemocap/pyproject.toml`, which makes sense because all of the poetry stuff is in `poetry_pyproject.toml`. It looks like `pyproject.toml` is a protected name, so ideally we can switch to having one pyproject that does it all. 

I think we can still merge main into development, since this action doesn't run on development, as it broken on the main branch anyways. Just wanted to make note of it as I notice it.

philipqueen:
Well, this didn't do the job somehow, as `development` is still showing up as 2 commits behind main

PR #510, opened 2023-12-13
Development 

jonmatthis:


philipqueen:
I looked into why this branch is failing the nox tests. It seems to be disk space issue for the last two tests (python 3.11 and coverage). This is likely an issue on github's side of things, as none of our packages are very big. If it continues to be an issue, there are some actions like https://github.com/jlumbroso/free-disk-space that could help, but for now I doubt it will continue to be a problem

philipqueen:
I've changed skellytracker (and updated the version) to turn off YOLO cropping as the default

pyproject.toml
```diff
@@ -80,6 +80,7 @@ dependencies = [
     "ipykernel==6.23.1",
     "plotly==5.14.1",
     "pydantic==1.*",
+    "skellytracker==2023.12.1006",
```
philipqueen, line 83: @jonmatthis do we want to be pinning these subrepos to precise versions (i.e. with `==` rather than `>=`)? In theory we're in control of any breaking changes.

philipqueen:
@aaroncherian can you test the most recent PyApp installer in the virus scan website you used? I'm curious if the opencv conflict stuff escalated our "threat" in the virus scanners eyes.

philipqueen:
The version testing failed due to a memory issue on the runner, not due to any code problem (it made it through fine to the coverage). I've rerequested that test, but all the python versions passed.

jonmatthis:
@philipqueen - Hit a unfamiliar `mediapipe` looking error when running on my Intel Mac (which I really need to do more often!!)

Does this look familiar? 
 
[message.txt](https://github.com/freemocap/freemocap/files/14397170/message.txt)

philipqueen:
> @philipqueen - Hit a unfamiliar `mediapipe` looking error when running on my Intel Mac (which I really need to do more often!!)
> 
> Does this look familiar?
> 
> [message.txt](https://github.com/freemocap/freemocap/files/14397170/message.txt)

No but it looks like this popped up for someone else very recently with 10.10: https://github.com/google/mediapipe/issues/5168
Do you have 10.10 or 10.9 installed?

philipqueen:
I was able to replicate it by upgrading to 10.10, which is just a few days old. I'm going to pin the version to under 10.10 for now.

Also we're still getting memory errors on testing, so I'm going to split the coverage testing out from the version testing to reduce the load on the runners.

jonmatthis:
> I was able to replicate it by upgrading to 10.10, which is just a few days old. I'm going to pin the version to under 10.10 for now.

Downgrading to `0.10.9` fixed it for me. 


> Also we're still getting memory errors on testing, so I'm going to split the coverage testing out from the version testing to reduce the load on the runners.

Incidentally, I started playing around with unittests in `skellycam` and had a few lightbulb moments - there's a pretty nice workflow where I just give botto the whole `.py` file and ask it to make a test for specific features. Its an easy way ti incrementally build up very thorough test coverage, and could be taught to lower-XP folks as a way to start contributing in a useful way that's unlikely to cause any collisions. @MocapMike might be into trying this out - we can chat about it later!

jonmatthis:
@philipqueen ok, all looks good to me - good to merge and update `pip`?

jonmatthis:
@philipqueen - another annoying bug - The GUI seems to freeze when the folder is processing now? I don't think it used to do that, but I can't be sure. 

Can you check on that and see if there is a quick fix? Maybe something in the "processing" animation in the log view that's not appropriately sheilded behind threads and whatnot? 

If there's not a quick/obvious fix, we can push this thorugh and fix that issue in a patch update 

(note to us: we should be making more `patch` updates, which would partially alleviate this annoying "big commit" stuff - lessons learned!)

jonmatthis:
> @philipqueen - another annoying bug - The GUI seems to freeze when the folder is processing now? I don't think it used to do that, but I can't be sure.

On closer inspection, its not the log updates - the GUI seems to freeze for *part* of the processing, but not all? Still confusing

jonmatthis:
#547 - new bug to fix, not a blocker for this push

philipqueen:
@jonmatthis I was able to download the most recent PyApp installer, process the sample data, and manipulate the gui without it blocking.

jonmatthis:
#548 - Annotated videos fail to produce when YOLO is on - We can still push this update, but this needs to be solved as ASAP as possible

jonmatthis:
> I was able to download the most recent PyApp installer, process the sample data, and manipulate the gui without it blocking.

Yeah, its something subtle - it definitely gets clunkier when things are running, but likely not worth a fix. With any luck this kind of thing will be obviated by updating the base architecture

mohdsha06:
i want to replace the skelton in app ui with different img.(for personal use only,college project).how to change skins color and text of the software.?

PR #514, opened 2023-12-20
Py App Installers

### LINKED ISSUE #415: PyApp - Single-click installer
Need to develop some way to install and run freemocap without the command-line

philipqueen:
Use PyApp to create installer executables to streamline the installation process for users. Rather than bundling the whole package into a binary like PyInstaller, PyApp creates a minimal exe (about 4mb), that will create an environment and download Freemocap the first time it is run, and future times it will open freemocap in the existing environment.

Closes #415 

This involves three different github actions, one for each OS we want to support. Details on those follow:
1. Linux
   - This one is the easiest. The pyapp binary is created, it's made executable by running `chmod -x`, and then zipped into a `.tar.gz` file. This should be unzippable natively on linux and then runnable.
   - Note: to remain an executable when downloaded by a user, it must be packaged as `.tar.gz`, otherwise it loses file permissions and cannot be run without resetting those permissions manually: https://github.com/actions/upload-artifact?tab=readme-ov-file#permission-loss
   - [x] Tested on Linux (by Trent)
2. Mac
   - The Mac code is nearly identical to the Linux code, although I've also added a step to bundle it into a mac app. This means it has the freemocap icon, and generally looks like a standard application.
   - One great thing with pyapp is the Mac app created on a non-m-series runner works fine on an m-series Mac.
   - The big hurdle with Mac apps is security/permissions. You can download and unzip the app, but when you do it prompts you that it isn't safe and asks if you want to move it to the trash. This is obviously undesirable, although by going to the settings you can allow it. We can circumvent this by "code signing" and maybe also "notarizing" the app, which requires $100/year apple developer account, but can be done (I believe) on github actions. Here are two links for how this process is done: [Creating Distribution-Signed Code for Mac](https://developer.apple.com/forums/thread/701514#701514021) and [Packaging Mac Software for Distribution](https://developer.apple.com/forums/thread/701581#701581021)
   - [x] Tested on Mac (Monterey M series)
   - [ ] Tested on Mac (Catalina M series) (@trentwirth and @MocapMike can help with this)
   - [ ] Tested on Mac (non-m-series)
   - [x] try to get it to open a Terminal window (not possible, see comment below)
   - [ ] Decide if we want to code sign and potentially notarize as well
3. Windows
   - The Windows version is the most different, because it is creating a `.exe` file. I don't know how this reacts with the github permission lost mentioned above. It is downloaded as a .zip file, but I don't have a windows machine to test on to see if the executable runs.
   - [x] Test that `.exe` is able to run on Windows
   - [x] Figure out how to get an icon for windows app (or see if correctly opens with one already)
   - [ ] Tested (for functionality) on Windows


.github/workflows/windows_installer.yml
```diff
@@ -0,0 +1,61 @@
...
+          cd $WORKSPACE/pyapp-v0.12.0
+          export PYAPP_PROJECT_NAME=freemocap
+          export PYAPP_PROJECT_VERSION=1.0.33
+          export PYAPP_PROJECT_DEPENDENCY_FILE=$WORKSPACE/requirements.txt
+          export PYAPP_EXEC_SCRIPT=$WORKSPACE/freemocap/__main__.py
+          cargo build --release
+          cargo install pyapp --force --root $WORKSPACE
+          cd $WORKSPACE
+          mv $WORKSPACE/bin/pyapp freemocap_app && chmod +x freemocap_app
+          tar -czvf freemocap_app.tar.gz freemocap_app
```
jonmatthis, line 56: Windows Installer creates a `.tar.gz` file - This like looks similar to the Mac/Linux builders. Do you know how to fix?

jonmatthis (changes requested):
Windows Installer  outputs a `..tar.gz` file - should that be a `.exe` or something?

.github/workflows/windows_installer.yml
philipqueen: Yep, the reason for that is here: https://github.com/actions/upload-artifact 
"Permission Loss
File permissions are not maintained during artifact upload. All directories will have 755 and all files will have 644. For example, if you make a file executable using chmod and then upload that file, post-download the file is no longer guaranteed to be set as an executable.

If you must preserve permissions, you can tar all of your files together before artifact upload. Post download, the tar file will maintain file permissions and case sensitivity."
So you should be able to use an upgraded zip manager like winRAR or 7Zip to unzip it. I'm not able to test it on windows, but we may have success just changing the file extension and uploading it as an `.exe` though, I'll try it out and ping you when it's uploaded and ready to test.

philipqueen: @jonmatthis the .exe build just finished, you should be download and check if it is able to run: https://github.com/freemocap/freemocap/actions/runs/7388774233

philipqueen:
I did some research to see if the bundled Mac app can open a terminal window to display our logging (stdout calls). Apparently, the Mac app launcher `Launch Services` writes all of std.out to `dev/null`, so we're out of luck there.
We could release two Mac versions (a bundled app, which is easier to install but doesn't have the progress bars, and a unix executable that shows the progress bars in a terminal window but is harder to set up). That combined with continuing to improve our logging might be the best solution.

philipqueen:
We may run into some problems with Mac security when users download the mac app. There's some good instructions here on how to navigate those issues:
https://www.darktable.org/install/#macos

jonmatthis (approved):
Super stoked to roll this out :D

philipqueen:
Redownloading the latest version of this release, it looks like PyApp is able to find a cached version of FreeMoCap from a prior update. Which is fun and all, except it means the code isn't updating when I redownload the app. I'll have to dig in to whether it's possible to prevent that without going into the depths of PyApp stuff. 
Since most users won't have downloaded a version of the app before, it shouldn't be an issue now, but it's definitely something we want figured out before our NEXT release. Ideally this one too, since the likelihood we'll want to patch this version is probably high

philipqueen:
> Redownloading the latest version of this release, it looks like PyApp is able to find a cached version of FreeMoCap from a prior update. Which is fun and all, except it means the code isn't updating when I redownload the app. I'll have to dig in to whether it's possible to prevent that without going into the depths of PyApp stuff. Since most users won't have downloaded a version of the app before, it shouldn't be an issue now, but it's definitely something we want figured out before our NEXT release. Ideally this one too, since the likelihood we'll want to patch this version is probably high

Never mind on this, the issue seems to be related to the version number. If we set a new version string in the github actions, PyApp will redownload the repo. This brings up a good point though - how do we get bumpver to update the version strings in the "Create Installer" actions?

jonmatthis:
@philipqueen Running this on my Windows machine, it did the terminal thing saying it was installing, then it closed without warning and nothing happens after that.

Clicking the app again pops a terminal which immediately closes

@aaroncherian - did you say it works on your PC?

philipqueen:
> @philipqueen Running this on my Windows machine, it did the terminal thing saying it was installing, then it closed without warning and nothing happens after that.
> 
> Clicking the app again pops a terminal which immediately closes
> 
> @aaroncherian - did you say it works on your PC?

How long did you wait? Aaron had a similar problem where he assumed it wasn't downloading, tried reopening it, and messed with the install/

jonmatthis:
Windows is killing it because it thinks its a virus :-/

image

PR #516, opened 2023-12-22
Fix development runners

philipqueen:
The nox files are failing on development, now with a new issue. It seems the disk freeing action messed with the libegl downloads enough that our current downloading of them wasn't enough to get opencv to run. I'm going to see if removing 3.8 testing freed enough disk space for the tests to run, and if not I'll debug some of the options in the free disk space workflow to see what we can get away with removing.

### DIFF
--- a/.github/workflows/version-testing.yml
+++ b/.github/workflows/version-testing.yml
@@ -12,6 +12,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


philipqueen:
Was able to get nox to pass by turning off these options in the "free disk space" workflow: large-packages, docker-images, swap-storage. This removes android, haskell, and .net data, while preserving whatever libegl packages we need to run opencv

PR #521, opened 2023-12-27
Change multiprocessing parameter options

### LINKED ISSUE #513: [UX/UI] Put "number of processes" selector behind a "use multiprocessing" boolean
I know what it's doing, and even I find this confusing: 

image

----

Basically, just need to recreate the previous "use multiprocessing" boolean checkbox, and then enable/disable (NOT hide/show) the "number to use" box to its state.

philipqueen:
This will fix #513, once we get consensus on how we want it set up.

Right now, I've changed "Number of Processes to Use" to "Max Number of Processes to Use", to make it more clear that we won't necessarily spin up every process available. I've also added limits to the parameter group, so the user can't select less than 1 process, or more than the number of cores they have available minus one. 

It's possible to implement different naming, i.e. have 1 show up as "1 (no multiprocessing)" or the max show as "MAX" by implementing the parameter as a group and creating a list that the parameter pulls from. Since this would involve some logic and we haven't decided on it, I haven't added it yet.

The same goes for adding a "Use Multiprocessing" bool that supersedes the "Number of Processes" - we can do it, but it will add quite a bit of gui logic depending on how we implement it, so I want to make sure it's the direction we want to go.

### DIFF
--- a/freemocap/gui/qt/widgets/control_panel/process_mocap_data_panel/parameter_groups/create_parameter_groups.py
+++ b/freemocap/gui/qt/widgets/control_panel/process_mocap_data_panel/parameter_groups/create_parameter_groups.py
@@ -42,7 +42,7 @@ RUN_3D_TRIANGULATION_NAME = "Run 3d triangulation?"
 
 RUN_BUTTERWORTH_FILTER_NAME = "Run butterworth filter?"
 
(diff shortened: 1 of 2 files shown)


jonmatthis (approved):
Looks good enough for me :)

PR #524, opened 2024-01-01
PATCH: logging to wrong location

philipqueen:
Currently, `configure_logging.py` imports the function that gets the path to log to form skellycam instead of freemocap, meaning freemocap logs are logged to the `skelly-cam-recordings` folder instead of the `freemocap_data` folder. This is obviously confusing to users, and makes it ahrder for us to get good log files while providing technical support. 

This PR changes the source of the `get_log_file_path` function from skellycam to freemocap, making logs save to `freemocap_data` again.

### DIFF
--- a/freemocap/system/logging/configure_logging.py
+++ b/freemocap/system/logging/configure_logging.py
@@ -4,7 +4,7 @@ from datetime import datetime
 from enum import Enum
 from logging.config import dictConfig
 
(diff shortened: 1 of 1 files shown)


jonmatthis:
Lol, good find! I knew there was something screwy going on with our Logs!

jonmatthis (approved):

PR #525, opened 2024-01-05
Log progress indicator

philipqueen:
This PR adds a progress indicator to the Log View Widget that shows when processing intensive tasks are in progress. This has been an ask for a while, and the primary motivator is to allow users of the Mac app bundle to see progress, as the app bundle does not allow logging to the terminal.

https://github.com/freemocap/freemocap/assets/24758117/d99c6271-67d9-4a6f-9f51-8fb9ee121faa

Limitations: It only indicates that the process is in progress, not what percentage/amount of the task is done like in a traditional progress bar. This is to limit the data that needs to be sent between processes (and the amount of quite technical code that would need to be replicated to get the progress bars to display correctly).

The other current limitation is it does not log during the blender processing. I intend to do a bit of a deep dive into how we log the blender output, as that is a current painpoint. The main thing that would need to happen is the log_view_widget would have to be passed to the export blender functions. However, currently Blender is blocking the GUI thread, so that would need to be fixed first.

### DIFF
--- a/freemocap/__init__.py
+++ b/freemocap/__init__.py
@@ -11,7 +11,7 @@ __repo_issues_url__ = f"{__repo_url__}issues"
 
 from freemocap.system.logging.configure_logging import configure_logging, LogLevel
 
-configure_logging(LogLevel.INFO)
+configure_logging(LogLevel.DEBUG)
 import logging
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


philipqueen:
@jonmatthis could you take a look at this? It's just merging into the logging branch for now so it doesn't technically need review, but I want to make sure it's a suitable solution

jonmatthis:
I like it!

PR #531, opened 2024-01-14
Put blender in threadworker

philipqueen:
This PR switches the Blender subprocesses to run form a threadworker. While `subprocess.Popen` is supposed to be non-blocking, it was blocking the main gui process. This allows the gui to run freely while processing, which is good in itself and also allows us to update the gui with a progress indicator while blender is processing.

The only negative change so far is that when Blender opens, it pops up a "Do you want to save this file?" dialog box, like it is exiting another blender window. Choosing either option seems to work fine. I think it's worth it to not block the gui, but @jonmatthis you might be able to figure out why this is happening.


jonmatthis (approved):
Nice!

Not sure about that "save this file" dialog thing, I'll take a look

PR #536, opened 2024-01-23
Generalize mediapipe code to other trackers

philipqueen:
Currently, we define a lot mediapipe specific variables that are passed around Freemocap. This is things like the number of points tracked, joint connections and names, and center of mass related numbers. The goal is to have all of these model specific variables be housed in Skellytracker, and imported into the `ProcessingParameterModel` in Freemocap. Then all references to mediapipe specific code can instead pull generalized variables from the parameter model.

There are a few places that are still mediapipe dependent, including:
- DataLoader class, which is only used in the DataSaver right now
- Blender Megascript Take 2 (which we may or may not find worth updating)

We also need to update the various subrepos, which are using their own copies of the mediapipe variables. The goal is to have these passed into the subrepos, which will also prevent a lot of existing duplication:
- Skellyviewer
- Skellyforge
- BlenderAddOn (we'll do this last, as there may be a lot of mediapipe dependency here)

This PR depends on a PR in skellytracker which has not yet been merged:
https://github.com/freemocap/skellytracker/pull/15
To get the latest skellytracker version, run `pip install skellytracker@git+https://github.com/freemocap/skellytracker@philip/flesh_out_mediapipe_model_info`


philipqueen:
Branches referenced:
https://github.com/freemocap/skellyforge/pull/3
https://github.com/freemocap/skelly_viewer/pull/10

jonmatthis (approved):
woo!

PR #539, opened 2024-01-31
Remove opencv import if conflicting dependencies exist

philipqueen:
We currently have conflicting dependencies between YOLO and MediaPipe, where the former downloads `opencv-python` and the latter `opencv-contrib-python`. This leads to errors when we try to run camera calibration in Anipose due to namespace clashing. Currently, there's no way to handle this clash in the `pyproject.toml` ecosystem, and there doesn't seem to be a resolution coming soon.

This PR adds a check in the `main` function in `__main__.py` to see if both opencv versions are installed in the current environment. If both are found (which should happen the first time freemocap is run after downloading), it runs a subprocess to pip uninstall both `opencv-python` and `opencv-contrib-python`. If the uninstall step was successful, it then installs `opencv-contrib-python`.

I don't like this solution at all, but it does work. My dislikes are:
1) We're changing the user's environments. Yes, we tell them to create a new environment, and if they're installing through PyApp it should be isolated in a venv anyways, but it's inevitable someone will be surprised by us uninstalling opencv and likely they won't be happy.
2) We have to uninstall both packages. Unfortunately, uninstalling `opencv-python` seems to mess with the namespace of `opencv-contrib-python`, so you have to uninstall both versions and reinstall just contrib for the fix to work. This adds some install time overhead, and has the possibility of uninstalling both opencv versions, but then failing to install for some reason, and leaving the user with neither opencv version.

Not sure of a better way than this for now, unless I'm able to figure out a solution with YOLO to make the opencv distribution editable somehow. But my guess is that change will take time, so for now this might be our best bet in integrating YOLO.

This still needs testing in the PyApp installer once we have it figured out in the normal GUI.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -14,6 +14,7 @@ per-file-ignores =
     ./freemocap/core_processes/process_motion_capture_videos/process_recording_folder.py:C901
     ./freemocap/core_processes/detecting_things_in_2d_images/mediapipe_stuff/mediapipe_skeleton_detector.py:C901
     ./freemocap/core_processes/detecting_things_in_2d_images/mediapipe_stuff/convert_mediapipe_npy_to_csv.py:E203
+    ./freemocap/gui/qt/widgets/control_panel/calibration_control_panel.py:C901
     ./experimental/batch_process/batch_process.py:E226
 # C901: function is too complex
 # E203: whitespace before ':' (PEP8 says this doesn't apply to slicing)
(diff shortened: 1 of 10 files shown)


philipqueen:
@jonmatthis you should take a look at this when you have time

jonmatthis (approved):

philipqueen:
@jonmatthis I've made this a dialog that opens when you open freemocap (after the welcome screen dialog). Feel free to change the text

Screen Shot 2024-02-05 at 8 49 50 PM

philipqueen:
Screen Shot 2024-02-07 at 10 41 50 PM

I changed the button text and style, let me know if you'd like more changes @jonmatthis

PR #543, opened 2024-02-23
PATCH: calibration not updating on success

philipqueen:
The calibration path is not being set properly after a successful calibration. This PR passes the calibration path explicitly into the `update_calibration_toml_path` function to bypass the control logic when `None` is passed in.

Right now, its set so the path is set to the toml path from the calibrations folder, but we can change that to the recording folder toml or the most recent calibration toml, whichever is best.


aaroncherian:
Yeah I gave it a shot and it worked for me!

jonmatthis (approved):

PR #566, opened 2024-03-20
get rid of blender 4 exclusion

aaroncherian:
Simple PR to get rid of the blender 4.0 exclusion in blender exe finding


philipqueen:
Looks good to me, it appears to just revert the prior change

jonmatthis (approved):

PR #567, opened 2024-03-20
Add YOLO Crop Parameters to Parameter Tree

philipqueen:
Parameterizes the YOLO crop method with model size, bounding box buffer method, and bounding box size. This will give the user more control over the process.

In the gui, it looks like this, with the entire yolo crop section being collapsible
Screen Shot 2024-03-20 at 4 02 55 PM


jonmatthis (approved):
Can we sneak, like, a `?` button that links to the `docs` somewhere into the `Process Data` panel? 

Esp bc/confirming - @philipqueen you said tooltips have never worked for you?

philipqueen:
> Can we sneak, like, a `?` button that links to the `docs` somewhere into the `Process Data` panel?
> 
> Esp bc/confirming - @philipqueen you said tooltips have never worked for you?

Tooltips show up, barely visible, only when I hover directly over the parameter box, but not on the parameter name. So they technically show up, but you would have to chance into them. I'll try to find some way to link to the docs there

PR #571, opened 2024-04-10
removing estimate segment lengths

aaroncherian:
Removing estimate skeleton segment lengths from the pipeline (we don't use the function anymore)


philipqueen:
This looks good to me!

For the stuff that was commented out, we can also just remove it. If we need it for any reason, it's in the git history

aaroncherian:
Oh yeah totally agree. Honestly the only reason was really for the laziness of my future self lol. All the commented stuff is `DataSaver` related - and I'm still unsure as to the role of `DataSaver` vs. the `save_data` function - but I thought there might be a chance that we'd want to do a 1 to 1 replacement of saving out the previous segment connections with the incoming segment connections. So I commented it out to easily bring it back if necessary without having to check the git history. Again, just laziness lol.

philipqueen (approved):

jonmatthis (approved):

PR #576, opened 2024-04-22
Update sample and test data

philipqueen:
This PR updates the sample and test data with new figshare links. The videos are the same, but the calibration has been rerun from the sample data, and gives better results than the original calibration we're currently including. It also adds functionality to download the sample data and test data into different folders, so both can be downloaded with overwriting each other (they both would download to `freemocap_sample_data` before). Finally, it fixes a name error in the error logging of the DownloadSampleDataThreadworker.

## To Test:
Switch to this branch, and select `download sample data` from the menu bar, and:
1) make sure it downloads successfully
2) make sure it is correctly set as the active recording
3) make sure it points to a calibration called `freemocap_sample_data_camera_calibration.toml`
4) For bonus points, process the data!
5) Repeat steps 1 through 4, but after selecting `download test data`

All of the above work properly on my machine.

Sample data before:
CurrentSampleDataOutput
Sample data with updated calibration:
NewCalibrationSampleDataOutput

### DIFF
--- a/freemocap/gui/qt/workers/download_sample_data_thread_worker.py
+++ b/freemocap/gui/qt/workers/download_sample_data_thread_worker.py
@@ -31,5 +31,5 @@ class DownloadDataThreadWorker(QThread):
 
         except Exception as e:  # noqa
             logger.exception(e)
-            logger.error(f"Error downloading sample data from {self._dowload_url}")
+            logger.error(f"Error downloading sample data from {self.download_url}")
             raise e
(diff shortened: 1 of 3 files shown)


jonmatthis (approved):

xiaoyeye111:
Please tell me how can I download sample data.I tried in GUI but it said "[ERROR ] [ProcessID: 19148, ThreadID: 3160] [freemocap.gui.qt.workers.download_sample_data_thread_worker:run():34] ::: Error downloading sample data from https://figshare.com/ndownloader/files/45797067.".It seems that it doesn't exist anymore

philipqueen:
Hi @xiaoyeye111, that link downloads properly for me (if you copy and paste it into a browser it should download). Is it possible you had a network problem? If you check the terminal/console window that you ran freemocap from, it may have a more detailed error message

xiaoyeye111:
> Hi @xiaoyeye111, that link downloads properly for me (if you copy and paste it into a browser it should download). Is it possible you had a network problem? If you check the terminal/console window that you ran freemocap from, it may have a more detailed error message

Okay I tried again and it downloads properly,thank you very much!

PR #579, opened 2024-04-25
Update subrepos to latest versions

philipqueen:
This PR brings in the most recent subrepo updates by specifying new versions in `pyproject.toml`.

It adds:
- Finding annotated videos first in skellyviewer
- Fixing default output and enabling rigify automatically in the blender addon
- Latest skellytracker (no specific update for freemocap, just a move of dependencies to extras)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,11 +68,11 @@ keywords = [
 
 dependencies = [
     "skellycam==2023.12.1090",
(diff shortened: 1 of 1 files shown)


philipqueen:
This is redundant with #570, so I'm going to close this to prevent some merge conflcits

PR #582, opened 2024-04-28
Patch: check for OSerror when accessing drives

### LINKED ISSUE #581: Start up fails on empty drives
It seems that freemocap (or rather `get_best_guess_of_blender_path.py`) doesn't know how to deal with empty drives, such as optical drives.

I had a user that had an empty blu-ray drive (F:) and `get_best_guess_of_blender_path.py` kept failing while trying to search that drive for `F:\Program Files\Blender Foundation`.

Problem solved by removing F from the list of drives to check.

Attached console log.
[New_Text_Document_6.txt](https://github.com/freemocap/freemocap/files/15142178/New_Text_Document_6.txt)

philipqueen:
Fixes #581. Checks for an OSerror when running `rglob` on a path. If `rglob` throws an OSerror it will return None, which will be handled by the outer `get_best_guess_of_blender_path()` function.

### DIFF
--- a/freemocap/core_processes/export_data/blender_stuff/get_best_guess_of_blender_path.py
+++ b/freemocap/core_processes/export_data/blender_stuff/get_best_guess_of_blender_path.py
@@ -8,7 +8,11 @@ logger = logging.getLogger(__name__)
 
 def guess_blender_exe_path_from_path(base_path: Union[str, Path]) -> Optional[Path]:
     base_path = Path(base_path)
-    blender_folder_paths = [path for path in base_path.rglob("blender.exe")]
+    try:
+        blender_folder_paths = [path for path in base_path.rglob("blender.exe")]
+    except OSError:
+        logger.info(f"Unable to access: {str(base_path)}")
+        return
 
     if blender_folder_paths:
 


freemocap/core_processes/export_data/blender_stuff/get_best_guess_of_blender_path.py
```diff
@@ -8,7 +8,11 @@
 
 def guess_blender_exe_path_from_path(base_path: Union[str, Path]) -> Optional[Path]:
     base_path = Path(base_path)
-    blender_folder_paths = [path for path in base_path.rglob("blender.exe")]
+    try:
+        blender_folder_paths = [path for path in base_path.rglob("blender.exe")]
+    except OSError:
+        logger.info(f"Unable to access: {str(base_path)}")
```
jonmatthis, line 14:
```suggestion
        logger.error(f"Unable to access: {str(base_path)}")
```

This should be logged as an error, right? or at least a warning?

jonmatthis (changes requested):

freemocap/core_processes/export_data/blender_stuff/get_best_guess_of_blender_path.py
jonmatthis: I'm trying to think of circumstances where this kind of error would arise - do you know any specifics?

jonmatthis (approved):
changing review to `approve` so you can merge on whatever decision you make on the log level

PR #588, opened 2024-05-14
Pydantic 2 Upgrade

philipqueen:
Update pydantic version from 1 to 2. Pydantic is already to 2.7, so this is overdue, but it's good to get the migration done before we add a lot of pydantic things for the FastAPI conversion.

## Subrepo PRs this depends on:
https://github.com/freemocap/skellycam/pull/57
https://github.com/freemocap/skellytracker/pull/35

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,35 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: "[BUG]"
+labels: bug
+assignees: ''
+
+---
+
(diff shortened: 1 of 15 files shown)


philipqueen:
The tests are failing because both skellycam and skellytracker depend on pydantic 1.*, so we're getting a dependency conflict. Once we update the subrepos we'll be able to update the versions for them and get the tests running

philipqueen:
Running this on my local environment with Pydantic 2.7 installed works well

jonmatthis (approved):
Just in time for `numpy 2.*` nonsense to begin 😌

PR #609, opened 2024-05-29
Support 3.12

philipqueen:
Adds support for [python 3.12](https://docs.python.org/3/whatsnew/3.12.html). 

## Why?

We've had an increasing number of users running into problems with trying to install in a python 3.12 environment, unwittingly installing an old release (from before we specified a max version), and getting roundabout errors. Supporting 3.12 will avoid this frequent error.

## Why Now?
We were previously blocked by a bug in mediapipe that didn't allow the holistic model to be run on Mac (https://github.com/google-ai-edge/mediapipe/issues/5168). This was fixed in 0.10.13, so we can now bump mediapipe versions and upgrade to supporting 3.12. 

## Subrepos
All of our subrepos also specify a maximum python version, and these will need to be updated. Currently the PRs just update the maximum version, but we may want to only specify a maximum version in freemocap itself and any suberepo with an explicit conflict, so we don't have to update everywhere every version bump of python.

The following PRs will need to be merged before this one:
https://github.com/freemocap/skellycam/pull/58
https://github.com/freemocap/skellytracker/pull/36
https://github.com/freemocap/skelly_viewer/pull/15
https://github.com/freemocap/skellyforge/pull/4
https://github.com/freemocap/skelly_synchronize/pull/34
https://github.com/freemocap/freemocap_blender_addon/pull/11


philipqueen:
This is still waiting on a new `skellyforge` version to be pushed to PyPI and updated in the pyproject.toml. After that, everything should be ready

philipqueen:
I've split the version testing into 2 different test sessions: one running 3.9 and 3.10, and the other running 3.11 and 3.12. This was primarily done to avoid the "out of space" errors we keep getting. This duplicates some setup code, but also allows the tests to run in parallel, reducing the overall delay between pushing commits and passing tests

jonmatthis:
> I've split the version testing into 2 different test sessions: one running 3.9 and 3.10, and the other running 3.11 and 3.12

I think this is fine for now - I bet we could find some other way of handling the 'out of memory' thing (maybe by cleaning up the cache between runs?). 

The fact that they run in parallel is a nice! It also makes me wonder if it is possible to  GH Actions  to run in parallel within a single workflow file?

jonmatthis (approved):

PR #611, opened 2024-06-11
Allow PySide 6.7 in pyproject

philipqueen:
PySide6 is now on version 6.7, and clicking through the GUI it appears no breaking changes have been made. This PR updates `pyproject.toml` so it allows both 6.6 and 6.7.

## To Test
`pip install git+https://github.com/freemocap/freemocap/@philip/update_qt_version` and play around in the GUI, clicking as much as possible

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,7 +80,7 @@ dependencies = [
     "ipykernel==6.23.1",
     "plotly==5.14.1",
     "pydantic==1.*",
-    "PySide6==6.6.*",
+    "PySide6>=6.6, <6.8",
     "packaging===23.2",
 ]
 


philipqueen:
Tests are failing because we're getting this error when trying to pip install on 3.9:
```
 ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
    unknown package:
        Expected sha256 7373479565e5bd963b9662857c40c20768bc0b5853334e2076a62cb039e91f74
             Got        7795ff16410b885ca04f5f4b3b6757698121166c37d7ab8569e859fc17ac306e
```
I haven't seen this before, and it installs + runs fine locally. It's building 6.7 for both. I'll look into this next week.

jonmatthis:
> we're getting this error when trying to pip install on 3.9:

I'm down to drop `3.9` support if you want

philipqueen:
@jonmatthis seems like it was just a temporary issue, tests are passing now

jonmatthis:
> seems like it was just a temporary issue, tests are passing now

@philipqueen that actually makes me feel LESS comfortable 😂 


There's no such thing as a temporary issue that goes away on its own, SOMETHING changed (or perhaps, there was some human error in the way the test were run). There's fundamentally no such thing as stochastic code (or I guess there is, but its called 'quantum computing')

its worth a poke to figure out what changed, but if nothing pops up we can just toss it in the 'discomfort-of-partial-knowledge' bucket and move on 😌

jonmatthis (approved):

PR #618, opened 2024-07-04
3d from Single Camera

philipqueen:
Adds a processing parameter to allow the user to decide whether or not to flatten the data from a single camera. Flattening is still the default, but this allows users to see what "3d from 2d" looks like. While the output is far from accurate, it's been requested often and may be of some use to animators.

I'm planning on sharing this PR to the discord to see if users who have requested this think the data is somehow usable/more desirable than the flattened data.

## To Download This Branch:

In a new environment, run `pip install "freemocap@git+https://github.com/freemocap/freemocap@philip/3d_from_single_camera"`. Then, running `freemocap` will bring up this branch.

Screen Shot of GUI option

### DIFF
--- a/.github/workflows/coverage-testing.yml
+++ b/.github/workflows/coverage-testing.yml
@@ -11,7 +11,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Free up disk space
         uses: jlumbroso/free-disk-space@main
         with:
@@ -21,7 +21,7 @@ jobs:
       - name: Set up Python 3.x
         uses: wntrblm/nox@2022.8.7
         with:
-          python-versions: "3.9, 3.10, 3.11"
+          python-versions: "3.12"
       - name: Install libegl1
         run: |
           sudo apt-get update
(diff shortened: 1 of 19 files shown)


jonmatthis:
Yeah, I think this is fine idea - I've seen others asking

Remember that there are 2 types of '3d from 2d data' (I think) - one of the outputs is 3d relative to the presumed camera position (I think using an estimate of mean human height, but not sure) and the other output has the 3d data in 'world' coordinates (unclear how that's calculated) 

They are probably the same data with some affine transformation applied, but we might want to give that extra level of choice and let uses pick which one to use. 

I'm half-remembering all of this btw, we can chat more about it later and/or check the mediapipe docs or BlazePose docs to see if they specify which is which

philipqueen:
@jonmatthis interesting, I'll have to look into that more when we add other 3d-from-2d capable trackers.

Mediapipe Holistic uses the camera depth with the hips as the origin, and scales it to roughly the scale of the x parameter (image width). It looks like the new Mediapipe Pose (which we're not using because they haven't updated Holistic yet) supports both Landmarks and WorldLandmarks, so we'll be able to support that choice when we upgrade to new Mediapipe

jonmatthis (approved):

PR #619, opened 2024-07-18
Update post_process_skeleton.py

aaroncherian:
Fix to make this branch of freemocap work with the `get_model_info_from_skellytracker` branch of `skellyforge`


philipqueen (approved):
Looks good to me! It's not formatted properly for Black but we'll fix that before merging to main

PR #621, opened 2024-07-19
Patch: Update Skellyviewer when processing finishes

philipqueen:
We haven't been updating skellyviewer when processing finishes. This PR adds the update call to the `_handle_processing_finished_signal` function in `MainWindow` so that skellyviewer will be updated after processing.


jonmatthis (approved):
Haven't tested this, but I trust you :)

PR #623, opened 2024-07-29
Remove residual confidence threshold cutoff

philipqueen:
Removes references to the `confidence_threshold_cutoff` in the GUI parameters and in the code. 

This parameter was part of the original reprojection filtering, which was removed in #373 and replaced with by camera reprojection filtering in #469.

It has no references in the code outside of the parameter model.


jonmatthis (approved):
:D

PR #658, opened 2024-12-12
Update addon version

philipqueen:
updates to the latest version of the Blender addon, and makes the required namespace changes to get it working.

### DIFF
--- a/.github/workflows/python-testing.yml
+++ b/.github/workflows/python-testing.yml
@@ -22,7 +22,7 @@ jobs:
         uses: actions/setup-python@v4
         with:
           # Semantic version range syntax or exact version of a Python version
-          python-version: '3.9'
+          python-version: '3.10'
           # Optional - x64 or x86 architecture, defaults to x64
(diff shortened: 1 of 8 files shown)


philipqueen:
It appears the Blender addon now specifies python>=3.10, but Freemocap allows >=3.9. Is there a reason we bumped the version in the addon?

jonmatthis (approved):
Looks good!

Gonna have to have a chat about the best way to de `ajc27_` every thing here  (@ajc27-git lol)

I don't think it's even possible to change the name of the pypi project so we may just need to make a whole new repo/project and deprecate this one.

I'm thinking we can call it `skellyblender` to make it consistent with the other sub-skellies ( also, it's a funny name lol)

We will need to do something similar for the current `skelly_viewer` which should be changed to `skellyviewer` to be consistent with the others

ajc27-git:
What was once an honor, now has become a burden. :sweat_smile:

Better to migrate the whole thing as you said.

Issue #653, opened 2024-11-18
[FEATURE REQUEST] Suggestions for improved data output


roaldarbol:
**Is your feature request related to a problem? Please describe.**
There are a few things I would love to see FreeMoCap do a bit better on the output side - I'll try to lay them all out here, but will be happy to split them up into separate issues to better track them individually.

1. Column naming is regex un-friendly
2. New output formats
3. Incorporate metadata

## Column naming is regex un-friendly
Column names may look like this, taken from the (proposed) docs:
> The column names look like `body_nose_x`, `right_forearm_y`, and `full_body_com_z`.

From these names, we can tell that the coordinate comes after the last `_`. Whereas the structure of `body_nose_x` seems to be `model_keypoint_coordinate`, `right_forearm_y` seems to be `side_keypoint_coordinate`, and `full_body_com_z` seems to be `model_stillmodel_keypoint_coordinate`. It would be *really* great to have the naming convention be generalised, making subsequent data wrangling easier. 
Given you output from multiple models simultaneously (I think, right? Both face and body in the same data for example?). I think I would prefer `model_keypoint_coordinate` - here `right_forearm` would simply become `rightforearm` (note it would already be much easier to regex because we know which word to split). 

For a more tidy syntax, I would recommend:
`time, model, keypoint,  x, y, z, reprojectionerror`

## New output formats
Currently you support `npy`, `csv` and `json`. It's great, as they're widely supported. However, they do have some drawbacks. `npy` is really Python-centric, and hard to work with elsewhere without interfacing with Python. `csv` lacks any metadata and the file size can become a problem. I don't have much experience with `json`, but I imagine it's similar to `csv`.

I would love to see two new formats supported:
1. Parquet
2. HDF5 / netCDF-4

**Parquet** is cross-language, widely supported and developed by Arrow. It compresses file size incredibly effectively and is great for long term storage. It also supports adding metadata.

**HDF5** (`h5`) is also supported in multiple languages and the multi-dimensional arrays make it a great option for storing movement data. Again, metadata becomes an integral of the data itself.

I can recommend following the issue over on movement, which has the purpose of facilitating better interoperability: https://github.com/neuroinformatics-unit/movement/issues/341'

## Incorporate metadata
[...]

PR #661, opened 2024-12-18
add tidy csv option

aaroncherian:
This PR is just to kick off some discussion on formatting for freemocap data, based talking to @philipqueen and the discussions in https://github.com/freemocap/freemocap/issues/653.

At the moment, this creates a new main csv file called `data_by_frame.csv` that has the same data as the existing `data_by_trajectory.csv` but in tidy format. I created a new csv instead of replacing the existing one just to make the changes as minimal as possible for this initial PR. And to be clear I'm not suggesting we merge this as it is - just using this as a starting point to look at some data formatting.


aaroncherian:
For reference, here's a snippet of what the tidied format looks like (csv is too big to attach).

image

philipqueen:
This looks good to me! I can see this being way easier to parse, and to generally use in downstream applications.

I'm surprised creating a list of dicts is fastest compared to directly appending to the dataframe, but some benchmarking on stack overflow does indicate that's the best method

philipqueen:
I've added a `model` category, and taken the prefixes like `body_`, `hand_`, and `face_`, and split them into the model (combined with the model name from the `ModelInfo`). We may end up wanting something like `model`(the name from `ModelInfo`) and `model_type`  (for body, hand, face, etc.), but I think this is sufficient for now.
Screenshot 2024-12-20 at 12 33 34 PM

aaroncherian:
Tested it on the freemocap sample data on my end and it looks good to me!

jonmatthis (approved):
Looks good!

PR #668, opened 2025-01-21
Reprojection error into tidy dataset

philipqueen:
As requested by @roaldarbol, adding the reprojection error into the tidy data csv output. 

## TODO:
- [ ] Right now there's a mismatch between the hand and face names that needs to be fixed, so only body points are showing up
- [ ] Decide if we want to include by camera


philipqueen:
For aniread purposes, the mean reprojection error is sufficient. So this should be ready to merge

PR #675, opened 2025-03-07
Patch: Fix Butterworth output not making it to final data

philipqueen:
A `hasattr` check on a dictionary was preventing freemocap from using the filtered data returned from skellyforge. Dictionaries don't have the keys as attributes, so it was consistently failing.

As a patch I've replaced the `hasattr` check with the best practice `in` check.

Note the `TODO` I added when this was merged though, that skellyforge should be deciding what its return data is, not freemocap.


aaroncherian:
I tested this on one of the treadmill sessions - using both our normal 7Hz cutoff and a 1Hz cutoff frequency on the BW filter. Seems to be working.
image

jonmatthis (approved):
ah, good find!

PR #676, opened 2025-03-10
Adding diagnostic steps to data processing pipeline

aaroncherian:
This PR is to get started on adding diagnostics to catch bugs that affect the quality of our output data.

The aspirational goal is to ensure that our data quality is always **monotonically increasing** across every version. 

---

We define "data quality" according the follow list of assumptions which define 'lower is better' error metrics: 

1) **Noise** - Assumes that oscillations above our butterworth cutoff frequency (10Hz) are caused my measurement error. Error metric: Butterworth residual (mean absolute value of  raw - butterworth filtered data)

2) **Smoothness** - Assumes that 'smooth' movement trajectories are are better, where 'smoothness' is defined according to a simple [jerk minimization] approach. Error metric: Mean jerk per frame (Triple derivative of position trajectories after filtering. NOTE - I wouldn't trust these `jerk` values for precise analysis, but I believe they should sufficiently track movement smoothness for our purposes) [TODO  - cite Flash & Hogan paper]

3) **Segment Rigidity** - Assumes the segments of the body should behave as rigid bodies connected by universal joints (aka the [Rigid Body Assumption](https://en.wikipedia.org/wiki/Rigid_body_dynamics)) Error metric: Mean residuals pre/post rigidification in skellyforge (TODO - link to the part of the code that does this)

---
The goal of this PR is to implement automations (via Github Action) to calculate these values for our standard Test dataset on each commit to the `main` branch, and raise a flag if the resulting error values go DOWN. 

This automation should help us ensure that the quality of our data increases monotonically (either stays the same or goes up, NEVER goes down). 

> *Better than its ever been, worse than it will ever be* 💀✨


jonmatthis:
@aaroncherian - Tweaked the top comment/description for this PR. Take a look and make sure it all looks right. Felt weird to edit a comment that you wrote, feels like I'm putting words in your mouth! 😂

aaroncherian:
Committed the first, very much in draft, version of a diagnostics test.

The current diagnostics pipeline works by:

1) Downloading the test data
2) Running calibration on the test data (**note**, not using the existing calibration toml in the sample data)
3) Running the test data through FreeMoCap
4) Taking the raw and processed 3D data and calculating the average total jerk and jerk/joint for each step
5) Comparing these values to values from previous versions of FreeMoCap that exist in CSVs saved within the `diagnostics` folder.
6) Formatting and uploading a nice html report as an artifact

The github action will run this pipeline on Python 3.11 across Windows, MacOS, and Ubuntu.

Things to note:
- I think I should double check my calculations for jerk. Currently I calculate jerk per dimension and average over X Y and Z to get the total, but there might be another/better way to do that
- The CSVs of previous version diagnostics are Windows specific - because as we've learned, there seem to be system specific differences, most likely due to something going on in `Anipose`. This needs further investigating. But the current diagnostics report for a system other than Windows will be more of a Windows vs. system comparison. 
- I added the recalibration step because at the moment we need that to generate a calibration per system so they all spit out the best data. However, that seems to like to (very slight) changes in our estimated values from run to run. We'll have to decide if we want to standardize those values with a specific calibration per system that we use - or if we accept the mild run-related variations. 

The system specific data variance is the current priority to look at. As a tldr, this action *does* work, and we could use it and just the Windows diagnostic report to tell if something has gone wrong with the pipeline if we wanted something committed faster. But there's still a lot of work to do here.

PR #677, opened 2025-03-10
Add data regression warning to welcome dialog

philipqueen:
Give a warning with the welcome screen dialog about the butterworth filter being skipped.

We can move this to a separate dialog if we want, but it will require updating the GUI state, so to get this done faster and less bug prone I just added it to the welcome dialog.

Screenshot 2025-03-10 at 11 39 50 AM


philipqueen:
ok, I've moved this to its own dialog that only shows up when the welcome screen doesn't show up. 


Screenshot 2025-03-10 at 12 05 32 PM

jonmatthis:
Prettied up the Warning a bit

image


Will merge and bump version when checks complete. Thanks!!

jonmatthis (approved):

Issue #681, opened 2025-03-19
[BUG] 3D data is different on the same test data depending on OS


aaroncherian:
We recently noticed, in the course of upgrading our diagnostic-running capabilities, that we appear to be getting different results on the test data between different OSs (the three in question being Windows, Ubuntu, and MacOS). This issue follows a deep dive into what the issues we've noted are, and progress at investigating what's causing them

## Current protocol for testing and running diagnostics is:
1) Use this creatively named fork I made of freemocap:
https://github.com/aaroncherian/freemocap_fork

2) There is a github actions diagnostic workflow that runs the diagnostics across Windows, MacOS, and Ubuntu 
Downloads the sample data from this Github release I made on the fork:
https://github.com/aaroncherian/freemocap_fork/releases/tag/v0.0.3-alpha
This sample data has no calibration toml included, just the videos
Runs the calibration
Runs the test data through FreeMoCap using that calibration
Calculates diagnostics (which is just the jerk data, at the moment) and compares them to previous versions of FreeMoCap (contained in csvs in the repo)
Diagnostics are located here: https://github.com/aaroncherian/freemocap_fork/tree/main/freemocap/diagnostics
Uploads an html report of the diagnostic data per system
Uploads the processed 3d data and calibration toml per system

Note - because the 'previous versions' of FreeMoCap diagnostics were calculated on my Windows computer - the Windows diagnostic report is the only true 'version vs version' comparison at the moment. The ones for the other systems, given the differences we see, will end up being more of a Windows vs. MacOS and Windows vs. Ubuntu comparison

## Differences we see in 3D data between OSs using the standard pipeline
First thing off the bat is a difference in the 3d data between OSs. I aligned the data from our three and plotted them. Code is here:
https://github.com/aaroncherian/freemocap_playground/blob/573f8e41697e7202c2a1c479ed08a9dafdd508bb/plotting/scatter_3d_system_test.py

![Image](https://github.com/user-attachments/assets/3d458922-a59f-46eb-a0b2-959cce4183e3)

You can see scaling differences between Mac and Windows, and big ones with Ubuntu

## Differences we see in calibration tomls between OSs using the standard pipeline
First one of note, which we've seen, is that the 'order' of the videos for Cam 0 1 and 2 are different across OSs.

For Windows its:

Cam 0: synced_video_1
Cam 1: synced_video_2
Cam 2: synced_video_3

On MacOS:
Cam 0: synced_video_2
Cam 1: synced_video_3
[...]

PR #686, opened 2025-04-17
Aaron/fix pin to 0 code

aaroncherian:
After a good bit of digging, I believe I've found the issue the `pin_camera_0_to_origin` code, which lies in how we're handling different coordinate systems.

Currently we're trying to move each cameras `tvec` by applying a transformation that would place camera 0's position at [0,0,0]. However, each cameras `tvec` exists in its own local reference frame, not in the world coordinate system. 

So currently we calculate a translation in the context of camera 0's coordinate system and then try to apply this same translation to other cameras with different orientations. This means that this function hasn't been rigidly moving the cameras as a set - it's been changing the spatial relationships between cameras.   This explains why different 'translations' that have been applied (for example, how which video was listed as 'camera 0' led to a different `translation to origin` vector) has led to differently scaled data.

The solution that I've tried to implement here is to calculate the translation needed to get camera 0 to the origin in the world coordinate system, not in camera 0's specific coordinate system. Then, we take this world-space translation, transform it into each cameras local coordinate system, and apply it.

### DIFF
--- a/freemocap/core_processes/capture_volume_calibration/anipose_camera_calibration/anipose_camera_calibrator.py
+++ b/freemocap/core_processes/capture_volume_calibration/anipose_camera_calibration/anipose_camera_calibrator.py
@@ -128,9 +128,10 @@ class AniposeCameraCalibrator:
             cam_group (freemocap_anipose.CameraGroup): A group of calibrated cameras whose extrinsics will be adjusted in place.
         """
 
-        self.align_rotations_to_cam0(cam_group=cam_group)
-        cam_group = self.shift_origin_to_cam0(cam_group=cam_group)
-
+        rvecs_new = self.align_rotations_to_cam0(cam_group=cam_group)
+        cam_group.set_rotations(rvecs_new)
+        tvecs_new = self.shift_origin_to_cam0(cam_group=cam_group)
+        cam_group.set_translations(tvecs_new)
         return cam_group
 
     def align_rotations_to_cam0(self, cam_group:freemocap_anipose.CameraGroup):
@@ -148,8 +149,7 @@ class AniposeCameraCalibrator:
             Ri_new,_ = cv2.Rodrigues(Ri @ R0.T)
             rvecs_new[i] = Ri_new.flatten()
 
-        cam_group.set_rotations(rvecs_new)
-        return cam_group
+        return rvecs_new
     
     def shift_origin_to_cam0(self, cam_group:freemocap_anipose.CameraGroup):
         # Get original translation and rotation vectors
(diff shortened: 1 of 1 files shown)


aaroncherian:
The big question, does it work? It seems to!

image

Here we have 3 plots - data with the original calibration (no pin to origin), with our current pin to origin code, and with the fixed pin to origin code in the PR. The fixed pin to origin and the original calibration now line up exactly

aaroncherian:
Another indication that it works:

Before this, we were seeing that the video chosen as `camera_0`, aka the reference frame, had a major impact on scaling:
image

But now the data is much more closely aligned regardless of which video was chosen as `camera 0`
image

aaroncherian:
What this also means is that the true issue causing the bugs mentioned in this issue (https://github.com/freemocap/freemocap/issues/681) was at the core, never related to the OS or the video order. The reason that when we sorted the videos we saw alignment is just because the `tvecs` for camera 0 were relatively the same between each OS, so the data looked really similar, but was not necessarily more accurate. They would have all suffered from the same 'distortion' from the improper pinning.

Thankfully, this also means that the video chosen as `camera 0` doesn't matter that much, which is a relief. But also, seems like we may have found something that's been affecting data quality for quite awhile, potentially.

aaroncherian:
I've added an extra step to zero out the rotation as well before zeroing out the translation, which seems to work?

image

freemocap/core_processes/capture_volume_calibration/anipose_camera_calibration/anipose_camera_calibrator.py
```diff
@@ -114,30 +114,51 @@ def calibrate_camera_capture_volume(self, pin_camera_0_to_origin: bool = False)
...
+
+        rvecs_new = np.empty_like(rvecs)
+        #rotates each cameras coordinate system into the new world space
+        #R0 maps world -> cam0, so R0.T (which is also the inverse) maps cam0 -> world
+        #R0 * R0^-1 = identity matrix, so this makes R0 the origin
+        # then we map every other camera into the new world space
+        for i in range(rvecs.shape[0]):
+            Ri,_ = cv2.Rodrigues(rvecs[i])
+            Ri_new,_ = cv2.Rodrigues(Ri @ R0.T)
+            print(i, Ri @ R0.T)
```
philipqueen, line 130: nit: we probably don't want this print statement

philipqueen (approved):
Running through a couple of test recordings this looks good!

jonmatthis (approved):
This is really great work, and such a perfect example of good tech sleuthing to find this unknown silent bug in our reconstruction process. You really pulled the thread from the original discovery (and fixing) of the OS-dependent video ordering bug, to the discovery of that reconstruction depending on the choice of Camera0 into this fix with both gives us a fully 6DoF-zero'd Camera0 as well as more accurate scaling. I think this was also (one of?) the first times you properly schooled me on an issue of basic 3d geometry, which was honestly extremely gratifying. A lot of great work for a 1-file-change PR 😂 

I'll merge this, and then figure out how to communicate changes in prep for a new minor release

aaroncherian:
Hey thanks for the kind words, I appreciate it. Glad that we were able to find an answer to some of the mystery scaling bugs that we had been seeing.

PR #687, opened 2025-04-28
detect variations of "Blender.app" on macOS

gonewest818:
On macOS the Blender application bundle can be relocated or renamed; for example maybe the user has multiple versions:

```
/Applications/Blender-4.0.2.app
/Applications/Blender-4.2.8.app
```

This PR uses a regex to accept any bundle containing the string "Blender" and ending with ".app"


jonmatthis (approved):
Awesome, easy fix! Thanks for the contribution!!

PR #688, opened 2025-05-09
QUICK FIX - Update CITATION.cff

trentwirth:
fixed broken zenodo link, fixed family/given name mix up for Trent & Endurance


jonmatthis (approved):
Good catch, thanks!

PR #695, opened 2025-06-03
use skellytracker for charuco detection in anipose

philipqueen:
Replaces charuco detection in Anipose with skellytracker's CharucoTracker. This gives us more control over the calibration pipeline, including getting parallel processing of the calibration videos. It also allows us to use the same detection for both the initial calibration and the charuco groundplane calculation.

II'll run through some tests of it, but the biggest thing I want to check for is error handling when there isn't a detection. As long as we don't regress there it should be good.

### DIFF
--- a/freemocap/core_processes/capture_volume_calibration/anipose_camera_calibration/anipose_camera_calibrator.py
+++ b/freemocap/core_processes/capture_volume_calibration/anipose_camera_calibration/anipose_camera_calibrator.py
@@ -21,7 +21,6 @@ from freemocap.utilities.get_video_paths import get_video_paths
 
 from freemocap.core_processes.capture_volume_calibration.anipose_camera_calibration.charuco_groundplane_utils import (
     compute_basis_vectors_of_new_reference, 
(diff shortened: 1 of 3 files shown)


philipqueen:
oh yeah that's been fixed. It was saving the `charuco_annotated_videos` one folder up in the tree, but I fixed that before merging on the skellytracker side

philipqueen:
@aaroncherian here's the diagnostic reports for the charuco changes from this into main: https://github.com/freemocap/freemocap/actions/runs/15739942121

aaroncherian:
> @aaroncherian here's the diagnostic reports for the charuco changes from this into main: https://github.com/freemocap/freemocap/actions/runs/15739942121

Yeah I was taking a look. All seems good - only interesting tidbit is that the mean estimated square size is the same across OS, which hasn't really happened before. I'm rerunning the action just to see if that's a coincidence or a feature of the new tracking.
image
image

aaroncherian:
Seems like coincidence!
image

Either way, I think this is good to go.

PR #420, opened 2023-06-06
Load sample data button

philipqueen:
This PR adds a "load sample data" button that allows the user to download a sample recording, and sets it as the active recording so it can be viewed or processed. Right now it's in the active recording widget.

Tasks:
- [x] Add a "Load sample data" button
- [x] make sure it's in a good place (easy to find, makes sense in the workflow)
- [x] Make it point to a good sample session, not the downsampled test session

Closes #392 
Sample Data Button

PR #704, opened 2025-06-18
Better ffmpeg missing behavior

### LINKED ISSUE #702: Add a checker for non-python dependencies
**Is your feature request related to a problem? Please describe.**
I've been trying to process pre-recorded footage (And it has to be pre-recorded because of issues with freemocap right now) and trying to get the dependencies I need to do that, just for synchronizing has been an absolute nightmare.

On my windows desktop I first ran into a bunch of errors and opened a bug report here. Turns out it's because FFMPEG isn't installed. I installed that, and now I'm running into completely different errors related to openCV. I figured ok, let me try it on my laptop, I just put OpenSuse Linux on there and it will probably have less issues, turns out there's even more. One dependency was missing, I could only find it because I searched github bugs, I'm having issues getting a specific version of FFMPEG now because the one in the repository doesn't have everything I need. I know that last one isn't on Freemocap, but its been non-stop frustration for me and this is cumulatively ruining my user experience.

**Describe the solution you'd like**
I understand that you can't make freemocap automatically install everything it needs without packaging it for individual linux distributions. But there should be some way for the program to run a check and say "FFMPEG not found" or "Library B not found". What's compounding the problem even more is the documentation doesn't list any of the required packages, so I have no manual way to check if I've got everything.

**Describe alternatives you've considered**
Issue #683 and making an official Flatpak would also solve this issue, and I'm in favor of doing it regardless. One it makes install easier if I don't have to mess with python environments. Two with SteamOS gaining popularity a lot of new linux users in the coming years will be starting with that (Or probably Bazzite) and unable to use freemocap as it is now. But given philipqueen's comment that sounds harder to pull off than this.

Alternatively as a non-feature solution, if there was just a list of exactly what's required in documentation for Windows and Linux.

**What is the use case for this feature?**
Anyone trying to install the program

EDIT:
The missing dependency in OpenSuse was libgthread

philipqueen:
This PR improves our handling of cases where FFMPEG is missing when the user tries to run synchronization. It adds a warning label in the synchronization dialog and adds a warning dialog if FFMPEG isn't found, and if the user still tries to synchronize it logs an error message and doesn't attempt the synchronization or import. It uses `shutil.which` to find if FFMPEG is available and on the path.

Screenshot of log error
Screenshot of dialog warning

Fixes #702


jonmatthis (approved):
Happy to approve preemptively, but could you go a few rounds with a bot to use some HTML and styling on this pop up to improve the visual hierarchy? 

Its never been a particularly aesthetic dialog, but with the addition of those new bullet points it starts to look like a floating soup of body text and its hard to tell what the important parts are. 

Check the `release_notes_widgets.py` and `release_notes_styles.py` for examples

In my experience, giving the bot a QWidget with simple prompts like "clean up the labels" and/or "adjust the styles to improve the visual hierarchy" tends to yield pretty solid single-shot results

Give it a shot and then go ahead and merge when ready!

jonmatthis:
And maybe add a link to instructions on how to install ffmpeg - many exist online, so pick your favorite 

We should always try to avoid instructions like "Install this thing and add it to your path" on the top level GUI - phrases like "add it to your path" can be infuriatingly opaque to the non-tech-trained

jonmatthis:
And at some point - Maybe enlist @aaroncherian to build more complete instructions into our GUI. 

Are there even any instructions around out there explaining what "Audio Cross Correlation" and "Brightness Contrast Detection" even means?

jonmatthis:
Looks great!

> I can link to this, or write something longer and add it to the freemocap docs and link to that, whichever you think is better.

This looks good, ideally it would in both the repo ReadMe AND the docs

In my mind, our repo ReadMe's should be more developer focused, with info including: 

- Summary (what it is and what it does)
- Installation (from source) 
- Basic Usage (including how you will know if it was successful)

We should generally assume that people reading the Repo docs have some basic technical know-how, and refer people to the full documentation for 'detailed instructions' 

The docs should be more for people who are using the software, specifically though the GUI. We should NOT assume any particular technical capacity and link to internal or external instructions on how to do basic stuff (like install ffmpeg, open a terminal etc) 

I wouldn't stress TOO much about getting all that aligned in this case, but we should try getting in the habit of making sure there's some lifeline link to instructional docs scattered through the variously confusing chunks of our repo 

( @aaroncherian this is kinda aligned with what I was saying about the new 'charuco groundplane' methods coming in from PR #691 - so might be nice to come up with some unified strategy for these two cases. I can imagine something taking inspiration from PR #420 but I'll leave specific implementation or decision to defer to y'all two)

jonmatthis:
@philipqueen - is this PR good to go? 

I know I dropped a bunch of other stuff about the ReadMe and docs on here, but that's more a brain dump than actual completion conditions for this particular task. Feel free to merge this in and defer the other tasks to the future if you want!

PR #707, opened 2025-06-23
Aaron/update to use skellymodels

aaroncherian:
Update to use skellymodels in FreeMoCap (and storing skellymodels in skellyforge)

To-do list:
[] Push new `skellyforge` package with `skellymodels` in it
[] Fix `ActiveRecordingInfo` 3d finding (need to have models save out `mediapipe_skeleton_3d.npy`, or some equivalent)
[] Update the freemocap blender addon

### DIFF
--- a/freemocap/core_processes/process_motion_capture_videos/process_recording_folder.py
+++ b/freemocap/core_processes/process_motion_capture_videos/process_recording_folder.py
@@ -118,12 +118,15 @@ def process_recording_folder(
             logging_queue.put(exception)
         raise exception
 
-    human = Human.from_tracked_points_numpy_array(
+    human: Human = Human.from_tracked_points_numpy_array(
         name="human",  # maybe there's a more useful name/identifier here?
         model_info=MediapipeModelInfo(),
         tracked_points_numpy_array=skel3d_frame_marker_xyz,
     )
 
+    human.put_skeleton_on_ground()
+    human.fix_hands_to_wrist()
+    
     try:
         human.calculate()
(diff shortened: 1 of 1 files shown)


philipqueen:
> @philipqueen So - one immediate issue is that the `ActiveRecordingInfo` check for whether data exists is failing on the 3D data check, because it's looking for a file that's labeled `{tracker_name}_skeleton_3d.npy`, which the model currently doesn't output. There's a variety of ways that we could go about this. We could have the model output a combined numpy array (which logically may make sense, since we output a combined csv and parquet). We could also run the check on a different file, if you think that would be better. Or a combination of both. Do you have opinions?

At this stage in the process I would lean towards having it output what freemocap is currently looking for. So. a combined numpy array (just make sure we combine it in the right order)

aaroncherian:
> > @philipqueen So - one immediate issue is that the `ActiveRecordingInfo` check for whether data exists is failing on the 3D data check, because it's looking for a file that's labeled `{tracker_name}_skeleton_3d.npy`, which the model currently doesn't output. There's a variety of ways that we could go about this. We could have the model output a combined numpy array (which logically may make sense, since we output a combined csv and parquet). We could also run the check on a different file, if you think that would be better. Or a combination of both. Do you have opinions?
> 
> At this stage in the process I would lean towards having it output what freemocap is currently looking for. So. a combined numpy array (just make sure we combine it in the right order)

Sounds good. The order is part of the input YAML to deconstruct the input numpy array, so I can just use that to reconstruct the output combined array.

PR #726, opened 2025-08-13
Expose multiprocessing to batch processing

philipqueen:
Default to using multiprocessing in the batch processing script, and expose the number of processes to users so they can control the max from the main block.

This should significantly speed up processing times for batch processing users.

### DIFF
--- a/experimental/batch_process/batch_process.py
+++ b/experimental/batch_process/batch_process.py
@@ -74,7 +74,7 @@ def process_recording_without_gui(
     else:
         logging.warning("No camera calibration toml file provided. May cause an error with multicamera recordings.")
 
-    recording_info_dict = rec.dict(exclude={"recording_info_model"})
+    recording_info_dict = rec.model_dump(exclude={"recording_info_model"})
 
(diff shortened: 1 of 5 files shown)


experimental/batch_process/batch_process.py
```diff
@@ -12,9 +15,10 @@
 
 
 def process_folder_of_session_folders(
-        path_to_folder_of_session_folders: Union[str, Path],
-        path_to_camera_calibration_toml: Optional[Union[str, Path]] = None,
-        path_to_blender_executable: Optional[Union[str, Path]] = None,
+    path_to_folder_of_session_folders: Union[str, Path],
+    path_to_camera_calibration_toml: Optional[Union[str, Path]] = None,
```
jonmatthis, line 19: Store paths as strings by default, and convert to Path when you need to. 

This lets us keep path data in a primitive (string) and also lets us get rid of a bunch of "Optionals" in our class/methods definitions.

jonmatthis (approved):
the comment about the str/path thing is mostly just to call it out, go ahead and merge this when its ready!

PR #733, opened 2025-09-01
Add Tiled PDF for Charuco Board (3×5 and 5×7)

PooyaDeperson:
This PR adds tiled PDF files for assembling Charuco boards at sizes 3×5 and 5×7


philipqueen:
Interesting approach to the charuco! 

The instruction sets are nice, but I think it could do with a little extra guidance. It should stretch that keeping everything aligned and perfectly matching the reference image is necessary to get a calibration. The charuco detection is ultimately finding the corners where the black and white squares meet, so having these corners on a perfect grid is necessary. This part `Ensure all pages are printed at the same scale and trimmed precisely as indicated.` is good, but we should just add onto it a note about precise assembly in addition to trimming. I think it's important to clarify that (probably better than I just did) to make sure people are aware that if they don't construct it correctly, the cameras might calibrate.

Also, what are the `timm_area_sample.webp` files? I don't get the purpose from the instructions.

Thanks for doing this! With a little more on the instruction front I'll be happy with it.

PR #776, opened 2026-03-26
Quality: Make camera probing tolerant of slow-start devices instead of hard-skipping at ~1s

### LINKED ISSUE #692: [BUG] It might be a virtual camera Skipping it.
Can you help me?
Camera 0 took 1.0007134999905247 seconds to produce a 2nd 
frame. It might be a virtual camera Skipping it.
[ WARN:0@27.159] global cap.cpp:480 cv::VideoCapture::open VIDEOIO(DSHOW): backend is generally available but can't be used to capture by index

huyhoang171106:
### Problem
The current detection flow is treating “slow second frame” as a virtual-camera signal and excluding the device too aggressively. This matches the user report where camera index `0` is skipped after ~1 second. The worker should stop using a single hard threshold and instead run a retry/warmup probe so real USB webcams (especially on Windows/DSHOW) are not dropped.

**Severity**: `high`
**File**: `freemocap/gui/qt/workers/cam_detection_thread_worker.py`

### Solution
Replace single-pass second-frame timing logic with a staged validation sequence: (1) open camera, (2) discard N warmup frames, (3) attempt frame reads for a configurable timeout window (e.g., 3–5s), (4) only mark as invalid if no valid frames after retries. If second-frame latency is high but frames eventually arrive, keep camera and mark it `slow_start=True` in metadata/logging rather than skipping.

### Changes
- `freemocap/gui/qt/workers/cam_detection_thread_worker.py` (new)

### Testing
- [x] Existing tests pass
- [x] Manual review completed
- [x] No new warnings/errors introduced

Closes #692

### DIFF
--- a/freemocap/gui/qt/workers/cam_detection_thread_worker.py
+++ b/freemocap/gui/qt/workers/cam_detection_thread_worker.py
@@ -1,0 +1,89 @@
+import logging
+import time
+from typing import Any, Dict, List, Optional
+
+import cv2
+from PySide6.QtCore import QThread, Signal
+
+logger = logging.getLogger(__name__)
+
+
+class CamDetectionThreadWorker(QThread):
+    in_progress = Signal(str)
+    finished = Signal(list)
+
+    def __init__(self, max_camera_index: int = 20, parent=None):
+        super().__init__(parent)
+        self._max_camera_index = max_camera_index
+        self._warmup_frames = 5
+        self._probe_timeout_seconds = 4.0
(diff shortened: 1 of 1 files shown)


tlobig:
I don't care much for all this AI chatter, but supporting slow starting cameras is something I would really need now. The low cost cameras I have are not properly detected under Linux, only Windows - which I don't care to use. I will see if I can hack my own local solution, but this feature seems like exactly what I need.

PR #793, opened 2026-06-09
Publish dummy skeleton when ring buffer doesn't have images

philipqueen:
This change fixed the hang in realtime. Essentially start up was too slow, and the frontend got stuck without an image. Passing a dummy skeleton instead of skipping helps it leave that state and lets realtime run properly


philipqueen:
A version of this got merged in already!

PR #791, opened 2026-06-04
Implement Pooyas UI

philipqueen:
Conversion of the front end code from the MUI version currently on development to one roughly matching the style laid out in the skellycam-ui, but with large improvements to make up for the differences between the apps

PR #797, opened 2026-06-13
Motion capture settings and others 

PooyaDeperson:
This `commit:` [prevent overflows on name dropdown selector](https://github.com/freemocap/freemocap/pull/791/commits/1e500a3abc264383cfdc1e12025a4311210601e6)
[1e500a3](https://github.com/freemocap/freemocap/pull/791/commits/1e500a3abc264383cfdc1e12025a4311210601e6) broken the dropdown interaction , made the dropdown a portal, visually it is okay now but cant select anymore within the modal, disppears, see video below

**I will bring it back to older state, for now and fix the dropdpwn issue in another way.**

image

`fixed in this commit`

broken dropdown interaction

--


PooyaDeperson:
So this PR covers the motion capture settings flow and how it would work in both streaming and playback.

The basic idea is that if you're in streaming mode and have Auto Process Mocap enabled, the Record Action button changes behavior a bit. Instead of immediately starting the recording, it first brings up the Motion Capture Settings for review. The user can quickly confirm the settings, and then recording starts. Once the recording is done, mocap processing runs automatically just like it does today.

Users can still go into Motion Capture Settings from the existing entry point we discussed before, but if Auto Process Mocap is enabled, we would still require a confirmation before starting the recording.

image

image


--


For playback, the flow is very similar. If a video is selected and the user wants to process mocap, we first show the Motion Capture Settings, let them review and confirm everything, and then start processing. I see this as the default behavior for playback since users are already explicitly choosing to process a video.

image


The reason I'm leaning this way is that these settings are pretty important and can have a big impact on the results. It feels better to put them in front of the user right before processing rather than assuming whatever was previously configured is still what they want. We can always add an option later to skip this step for power users, but for now I think it makes sense to have users at least glance at the settings and hit confirm before we kick off a process.



Streaming Mode (Auto Process Mocap ON)

Record Action
      │
      ▼
Motion Capture Settings
      │
   Confirm
      │
      ▼
Start Recording
      │
      ▼
Auto Process Mocap


Playback Mode

Select Video
      │
      ▼
Process Mocap
      │
      ▼
Motion Capture Settings
      │
   Confirm
      │
      ▼
Start Processing

PR #801, opened 2026-06-17
Dev -  freemocap v2 onboarding tooltips plus styles

PooyaDeperson:
image

- [x] this tooltip is good to go, and works.

hello @philipqueen 

these 3 tooltips need the proper logic . thank you 

1st show connecting... once connected
2nd, connect cameras
3rd, calibration.

image

added comment here  `BasePanelLayout.tsx`

image

---

TO DO:

Thanks, a few notes that somehow I believe got removed from the Skellycam to Freemocap transition:

- [x] **For me**, add the onboarding tooltips and visual guides.

..
.

I also need to redesign these:

- [ ] Progress modals (the ones at the bottom right)

- [ ] A new modal for mocap processing that can be ignored and minimized, but by default is a modal in the center of the screen. It should show the progress of what is happening behind the scenes and at what stage we are.

- [ ]  Playback and project stats, where we clearly show the project's statistics.

- [ ]  Recording finished confirmation modal, either for calibration or recording.

...
..
.

 Moreover, these are for the next stages on the purely for UX interactions.

Calibration Record and Calibration Flow:

1. Click "Record and Calibrate."

2. Record and calibrate: A modal confirmation appears before starting to record. It should say "Get set and ready," ask the user what to do and what to prepare, and explain what happens next.

3. Recording starts (we already have this state).

4. Recording finished (we already have this), but it also needs a minimum loading and pacing that clearly shows saving or loading. Sometimes, state changes that are too quick cannot be perceived by the user.

Calibration: Import Video and Import TOML:

1. Select "Import Video."
2. Show a modal about what to select, how, and where.
3. Folder selected: What's loaded and what has changed (we already have this state but it could be improved).

..

I also need to ask you, @philipqueen, to kindly 
- [ ] add the interactions on this explainer tooltip button in the calibration module here. It should show and hide the tooltip that is being shown.

also
- [ ] We need to work on combining the active recording with the playback.


philipqueen:
I think all the behavior here is correct! I'm going to go ahead and merge it, thanks!

PR #806, opened 2026-06-20
playback directory path aligns right and refresh button is grouped wi…

PooyaDeperson:
image


philipqueen:
looks good, thanks for fixing!

PR #834, opened 2026-07-29
Freemocap v2.00 alpha -  UI Alignment and Layout Improvements

PooyaDeperson:
- [x] Playback and Stream Mode grid settings flyout menus are now identical.

image

---

- [x] Splash modal buttons are now aligned.

image

---

- [x] Guided tooltip buttons are now aligned in a row and wrap into multiple rows when there isn't enough horizontal space.

image

---

- [x] logs terminal - badge text contrast fix

image

PR #835, opened 2026-07-29
Freemocap v2.00 alpha design updates

PooyaDeperson:
- [x] Playback and Stream Mode grid settings flyout menus are now identical.

image

---

- [x] Splash modal buttons are now aligned.

image

---

- [x] Guided tooltip buttons are now aligned in a row and wrap into multiple rows when there isn't enough horizontal space.

image

---

- [x] logs terminal - badge text contrast fix

image

### DIFF
--- a/freemocap-ui/package-lock.json
+++ b/freemocap-ui/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "freemocap",
-  "version": "2.0.0-alpha.1",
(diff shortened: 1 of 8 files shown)


PooyaDeperson:
iGNORE thiis PR - instead use  https://github.com/freemocap/freemocap/pull/834

thankyou

PR #837, opened 2026-07-29
uses segmented control ui - in mocap settings for detector and pose model size and also m…

PooyaDeperson:
image

image

--

adds padding

image


aaroncherian:
Looks good! Seems to work on my end. Next week I'll be able to run some full tests on the development branch as a whole.

Issue #849, opened 2026-08-05
[BUG] Post-processing Butterworth filter uses a hardcoded 30 fps regardless of recording rate


slrove:
**Describe the bug**
FilterConfig.sampling_rate defaults to 30.0 and is never set from the recording. The Butterworth low-pass is therefore designed for a framerate the capture never had, and the effective cutoff scales with the error:

`effective_cutoff = requested_cutoff * true_fps / 30.0`

**To Reproduce**
1. Record at any rate other than exactly 30 fps
1. Process with default filter settings (6 Hz, order 4)
1. Compare the requested cutoff with 6 * true_fps / 30

**Expected behavior**
The filter uses the recording's actual framerate.

**Environment Information:**
v2.0.0-alpha.20, Windows 11, 4 cameras

**Additional context**
Per-camera container fps can't be used to fix this: Four cameras in one synchronized recording could report 34.28 / 34.48 / 34.66 / 34.74 while sharing an identical frame count. The multiframe timestamps CSV is the only record of the shared rate.

davidbayus:
Thank you for filing this so clearly, and glad to see the fix land in #850 today. We hit the same bug in the 1.x line this summer, and since the fix is V2 side, here are our measurements and a 1.x workaround in case they are useful to anyone on the stable release.

In freemocap 1.8.2 the default comes from `post_processing_parameter_models.py` (`framerate: float = 30.0`) and is consumed as the Butterworth sampling rate in `post_process_skeleton.py`. The GUI Framerate field does pass through correctly, but it sits in a collapsed-by-default group preset to 30, and its own tooltip already flags the TODO to derive it from recorded timestamps, which is exactly what #850 now does.

Measured cost on a 60 fps six-camera take, for anyone wondering what this bug was worth: the intended 7 Hz cutoff behaves as roughly 14 Hz, so every take arrives half filtered. Re-processing the same videos with the same calibration and only the clock corrected cut above-8 Hz noise by about 90 percent at every landmark group we track (wrists 5.40 to 0.50 mm RMS, ankles 4.51 to 0.39, shoulders 1.05 to 0.10, hips 1.02 to 0.10), while the 0 to 6 Hz real-motion band moved about 0.01 percent.

[...]

PR #850, opened 2026-08-05
Make Blender export detector-aware and match the filter to real framerate

slrove:
Fixes #<issue 1>
Fixes #<issue 2>

- **#<issue 1>** — `REQUIRED_BLENDER_INPUT_FILES` was hardcoded to `mediapipe_*`,
  so an rtmpose recording (the default detector) always failed the pre-flight
  check and could never be exported. Now selects the required set from whichever
  detector's output is actually present, keeping #846's legacy-filename fallback.
  `REQUIRED_BLENDER_INPUT_FILES` stays as an alias so existing importers don't break.

- **#<issue 2>** — `FilterConfig.sampling_rate` defaulted to 30.0 and was never set
  from the recording, so the Butterworth low-pass was designed for a framerate the
  capture never had. Effective cutoff scaled with the error: 6 Hz requested became
  6.22 Hz at 31.088 fps, and would be 12 Hz at 60 fps. Now derived from the
  multiframe timestamps.

Per-camera container fps can't be used for this — four cameras in one synchronized
recording report 34.28 / 34.48 / 34.66 / 34.74 while sharing an identical frame
count. Only the multiframe timestamps carry the shared rate.

Falls back with a warning rather than substituting a plausible-looking default,
since silently guessing is what caused the original bug.


slrove:
> CI failure in test.yml is unrelated to this change — astral-sh/setup-uv@v4 isn't on the org's allowed-actions list. build-installers-pyinstaller.yml:197 already uses the pinned SHA cec208311dfd045dd5311c1add060b2062131d57; test.yml:38 was missed. Present on main, and this PR touches no workflow files.

The one-line fix if needed:

```
# .github/workflows/test.yml:38
uses: astral-sh/setup-uv@cec208311dfd045dd5311c1add060b2062131d57
```

philipqueen:
Thanks for the work on this @slrove - This is very helpful, both in fixing a bug in the filtering code and in prepping freemocap for non-mediapipe blender exports.

@jonmatthis this is ready for review, but depends on https://github.com/freemocap/skellycam/pull/141 to work. Once that's merged to main I'll do a final cleanup before merging this.

I changed a few things:
1. The filter config was correctly updated, but because it was made as a new reference it didn't update the actual filtering. I passed the new reference into the filter code instead of the stale one to fix it.
2. I made the system first check `timestamps/*_stats.json` for the mean framerate to get a reliable single number. This was broken before, so I had to make a change to skellycam to get it working. But because it reports exactly what we want, it's a clean way of getting fps.
3. I cleaned up the csv searching logic a bit, because we can expect the naming to be consistent.
4. Because the blender addon still can't process nonmediapipe trackers, I added some gating to prevent it trying to run in blender when rtmpose is used. We can remove this as soon as the blender addon is updated.

philipqueen:
Fixes #849

philipqueen:
@jonmatthis I saw you merged the skellycam change so I went ahead and did the final cleanup here (changing the skellycam branch reference and updating the lock file) - this should be good to review/merge when you're ready

freemocap-ui/src/components/control-panels/mocap-control-panel/BlenderSection.tsx
```diff
@@ -3,19 +3,25 @@ import {useBlender} from '@/hooks/useBlender';
...
 import IconButton from '@/components/ui-components/IconButton';
+import type {DetectorType} from '@/store/slices/mocap/mocap-slice';
 
 interface BlenderSectionProps {
     recordingFolderPath: string | null | undefined;
     disabled?: boolean;
     /** When provided, the "Open .blend in Blender" button is disabled unless true. */
     hasBlendFile?: boolean;
+    /** The freemocap_blender_addon only understands MediaPipe output so far. */
+    detectorType?: DetectorType;
```
jonmatthis, line 14: I think this is a fine enough way to handle/communicate the mediapipe->blender limitation for the time being. 

The near-term plan is to make Blender read  HumanActor formatted data (from skellymodels/skellycam), so the originating tracker will be less of an issue. 

So it will be `[Any SkellyTracker Output] -> [SkellyForge HumanActor] -> [Blender]`

freemocap/core/tasks/mocap/mocap_helpers/recording_framerate.py
```diff
@@ -0,0 +1,125 @@
+"""Derive a recording's true framerate from the timestamps written alongside its videos.
```
jonmatthis, line 1: Please double check this functionlaity isnt duplicated elsewhere (possibly in the `VideoGroup` object?) 

I can't remember if that object checks the timestamps or just the frame count. Worth sending through the machine with a note to check if this functionlaity already exists elsewhere. 

In any case, ideally this should get moved out of the `utilities` folder and moved somewhere appropriate in the existing file tree

jonmatthis (approved):
Generally looks good - I added a few minor comments through out. 

main thing is the recognition that this is a nice bandaid on the `mediapipe only` foible of our blender export, while the real fix will come when @aaroncherian  and I refactor the back half of our pipeline (skellyforge -> skellyblender) for the `v2` software (which will roughly correspond the to shift from `alpha` to `beta`)

freemocap-ui/src/components/control-panels/mocap-control-panel/BlenderSection.tsx
philipqueen: Great, this checks out with my understanding. It should be easy to remove this once the blender addon is ready, but in the meantime this should help prevent confusion from users I've been helping out with on discord

freemocap/core/tasks/mocap/mocap_helpers/recording_framerate.py
philipqueen: Great, I've moved it into mocap helpers where it fits better. It isn't duplicated elsewhere, so it does need to exist in some way.

PR #855, opened 2026-08-08
Fix tooltip images

### LINKED ISSUE #825: [BUG] Tooltip images aren't loading - freemocap v2. alpha windows cuda
**Describe the bug**

The tooltip images aren't loading. I think there might be an issue with the build, since this was working fine in my local development build when running from VS Code.

Tested on the Windows CUDA version.

**To Reproduce**
Open the app.
Go to the calibration section.
The tooltip images don't load.

Image

**Expected behavior**

The images should load normally, similar to the splash screen.

For reference, the splash screen uses a different method to load images (via the VSS background), which is also how the app icon is loaded. If that's the reason this works, we could use the same approach for the tooltip images.

philipqueen:
Fixes #825 - the tooltip images are showing up in the dev version but not the build version. This PR fixes it by using vite references to the images and importing them, instead of referring to them by a relative path that breaks in the build.


aaroncherian:
image

@philipqueen Great! Ran the installer action on this, and at the very least for the Windows one it works. I'll leave it to you whether you would like to test the linux/mac builds or go ahead and merge this into development. 

You can find the builds for this PR here:
https://github.com/freemocap/freemocap/actions/runs/31425820542

philipqueen:
I've confirmed on mac, so I'll go ahead and merge it

PR #857, opened 2026-08-09
Send websocket message when pipeline is stopped by user

### LINKED ISSUE #829: [BUG] Processing Remains Locked After Manual Stop ("Process Already in Use") -
**Describe the bug**

###  However, if I then try to process the same recording or another recording selected from the playback view, the processing modal displays the message "Process is already in use."

**Windows CUDA v2 Alpha 16** 

Image

Image

Image

philipqueen:
Fixes #829 - has the backend send a websocket message when the user stops a pipeline, mirroring the pattern when the pipeline finishes successfully on its own.

Screenshot 2026-08-08 at 7 36 46 PM


aaroncherian:
image

Great. Confirmed that a websocket message does show up.

PR #858, opened 2026-08-09
Hide ceres calibration until ready

philipqueen:
The pyceres based calibration method is currently untested, and we don't bundle pyceres with the app. This means there is a calibration method option where one option cannot run. This PR hides that option when ceres isn't available. This prevents confusion among users for now, and will show the option as soon as we choose to bundle ceres.


aaroncherian:
Tested it - no pyceres calibration shows up, anipose calibration works as normal.

PR #859, opened 2026-08-09
Import videos

philipqueen:
Add a button to import videos, matching previous functionality in freemocap. This copies the videos into a `{recording name}/synchronized_videos` folder. Right now it checks if the videos are synchronized, and if they aren't it blocks importing. I'm working on a rewrite of skelly_synchronize which should be done soon - we can either merge this as is and add synch later, or wait for the rewrite (with fastapi endpoints) to be done.


PooyaDeperson:
Oh, cool. So this is the same function as if you were going to playback mode and had selected the folder where the videos are, right?

So, in this case, the button will allow you to select directly from the splash screen, and once selected, if successful, it will send you to playback mode with that folder selected as active for you. It's like a shortcut.

I already have a design for this; I just need to implement it.

philipqueen:
Yes, similar, but the difference is this moves the videos into your recording directory, and puts them in the folder name freemocap expects. So similar to the function you described, but with some folder renaming and organization to help users

PR #885, opened 2026-09-01
Mediapipe Hand Detection Upgrade using ROI Cropping

ajc27-git:
Here is a proposed upgrade for the hand detection of the mediapipe detector.
In the new MediaPipe Tasks API setup, detectors run independently without a hierarchical cascade. As a result, HandLandmarker runs on the entire full-resolution frame. Because hands occupy a very small percentage of a full-body mocap frame, the detection rate drops significantly compared to v1.8.2 (which used MediaPipe Holistic). This leads to missing hand tracking and corrupted/invalid bone lengths.

This PR does the following:
- Modify mocap_task_config to execute the hand detector as a child stage of the body detector.
- Reorder the mediapipe marker columns because skellyforge expects [body, right_hand, left_hand, face].


aaroncherian:
Closing this, as your work was integrated into #890!

PR #888, opened 2026-09-11
Expose virtual cameras to users

philipqueen:
Add an option in "settings" to expose virtual cameras


aaroncherian:
Tested and works!

PR #893, opened 2026-09-17
Update build-installers-pyinstaller.yml

aaroncherian:
Testing to see if this fix solves an issue with CPU only computers still having GPU onnx and other gpu packages installed on it.


aaroncherian:
image

This seems to work properly. Going to merge it in.

PR #894, opened 2026-09-18
Add blender export options to GUI

ajc27-git:
This PR adds support for passing Blender export options from the GUI and applying them to the export pipeline.

Options added:

- 3D Model Formats: FBX and BVH
- Armature Rest Pose (useful for retargeting): T-Pose (default) or A-Pose
- Apply Foot Locking (disabled by default)
- Limit Hand Markers Range of Motion (disabled by default)

Note: I grouped the last two options under a "Motion Cleanup" subsection, but we can rename it to "Animation Motion Cleanup" or something similar to clarify that these are heuristic/stylistic adjustments rather than scientifically validated corrections.

image


philipqueen (approved):
This looks good to me!

The only question is whether we want to call `model_dump` with `exclude_unset=True`, which would use the blender addon default when a parameter isn't set. This might be useful if the addon api changes, or it might be a footgun for sending parameters that don't match. Approving this now in favor of the second option

